PR #44, opened 2019-12-09
Don't set full version of libtiledb soname

Shelnutt2:
We should not be hard coding the full soname version for libtiledb in native loader.


tdenniston:
I don't think this will work -- I believe only the explicitly versioned `.so` is copied into the `.jar` (and not the versionless symlink).

Shelnutt2:
> I don't think this will work -- I believe only the explicitly versioned `.so` is copied into the `.jar` (and not the versionless symlink).

I will test this some more, but in TileDB-Java we only set the `libtiledb`, https://github.com/TileDB-Inc/TileDB-Java/blob/6bef2b3667a03c8b5b1862d54178555c385225d8/src/main/java/io/tiledb/libtiledb/NativeLibLoader.java#L45

PR #59, opened 2020-03-02
More conda build fixes

ihnorton:
- Get libtiledbvcf from environment (PREFIX) if available
- Use -g and -O0 with debug option
- Don't overlink the arrow core library
- Bump the pyarrow minimum version to fix test errors

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -39,13 +39,18 @@ import shutil
 import subprocess
 import sys
 
+TILEDBVCF_DEBUG_BUILD = False
+LIBTILEDBVCF_PATH = None
+
 args = sys.argv[:]
 for arg in args:
   if arg.find('--debug') == 0:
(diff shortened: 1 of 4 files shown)


apis/python/setup.py
```diff
@@ -194,9 +207,10 @@ class BuildExtCmd(build_ext):
     """Builds the Pybind11 extension module."""
 
     def build_extensions(self):
-        opts = ['-std=c++11', '-g']
-        if not TILEDBVCF_DEBUG_BUILD:
-            opts.append('-O2')
+        opts = ['-std=c++11']
```
Shelnutt2: What's the reason for not including debug symbols in the non-debug build? I don't know how much symbols help python extension but for c++ libraries we generally don't strip the symbols out to assist in debugging.

ihnorton: Lost in shuffle between main and conda build trees. Restored, along with the `O2`.

PR #60, opened 2020-03-05
Add destructor for reader

Shelnutt2:
In the destructor we only make sure any async queries that are active finish before allow everything to be cleaned up. This prevent deleting the attribute/coords buffers from an active tiledb query.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -39,6 +39,16 @@ namespace vcf {
 Reader::Reader() {
 }
 
+Reader::~Reader() {
+  if (read_state_.async_query.valid()) {
+    // We must wait for the inflight query to finish before we destroy
+    // everything. If we don't its possible to delete the buffers in the middle
+    // of an active query
+    ctx_->cancel_tasks();
+    read_state_.async_query.wait();
+  }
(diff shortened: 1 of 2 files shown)


libtiledbvcf/src/read/reader.cc
```diff
@@ -39,6 +39,15 @@ namespace vcf {
 Reader::Reader() {
 }
 
+Reader::~Reader() {
+  if (read_state_.async_query.valid()) {
+    // We must wait for the inflight query to finish before we destroy
+    // everything. If we don't its possible to delete the buffers in the middle
+    // of an active query
+    read_state_.async_query.wait();
```
stavrospapadopoulos: That was a great catch. Perhaps in the future we can add cancelation capabilities (in case the query has to wait for a long time)?

Shelnutt2: I could add a `ctx_->cancel_tasks();` here. I've not used the cancelling of context tasks much, but since we truly don't care about the active query here we can call cancel and worry about making the behavior better in core tiledb over time?

stavrospapadopoulos: Good idea, it might be worth trying.

PR #66, opened 2020-03-25
Java API

gsvic:


apis/java/settings.gradle
```diff
@@ -0,0 +1,10 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * The settings file is used to specify which projects to include in your build.
+ * 
+ * Detailed information about configuring a multi-project build in Gradle can be found
+ * in the user guide at https://docs.gradle.org/4.8/userguide/multi_project_builds.html
+ */
+
+rootProject.name = 'tiledb-vcf-java'
```
Shelnutt2: should we drop the tiledb prefix here since its `io.tiledb.tiledb-vcf-java`? Also maybe we drop the `-java` too? Thoughts?

gsvic: @Shelnutt2 
Hm.  For the project name, I followed the same convention that we also use in spark
https://github.com/TileDB-Inc/TileDB-VCF/blob/master/apis/spark/settings.gradle#L10

In my opinion, I would keep the -java to discriminate it from the multiple VCF projects under the APIs dir. What do you think?

Shelnutt2: This project name is just used for importing the jar in gradle right?

gsvic: You mean, for example, importing the vcf-java jar to another project? If you mean that, yes, that's correct

Shelnutt2: Yes that is what I mean, thanks for the clarification. We can leave as is.

gsvic:
> Looks good. One question on the naming. Should we add some java tests in this PR also?

Totally agree. I will work on that, and I'll include a couple of tests into this PR.

PR #68, opened 2020-03-30
Validity Checks

gsvic:
Contains validity checks on
- Buffer Size (should be < 2048MB)
- Long to Int conversion

### DIFF
--- a/apis/spark/src/main/java/io/tiledb/vcf/VCFInputPartitionReader.java
+++ b/apis/spark/src/main/java/io/tiledb/vcf/VCFInputPartitionReader.java
@@ -282,8 +282,15 @@ public class VCFInputPartitionReader implements InputPartitionReader<ColumnarBat
       }
 
       if (bufferSizeMB > 2048) {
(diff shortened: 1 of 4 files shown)


apis/spark/src/main/java/io/tiledb/vcf/VCFInputPartitionReader.java
```diff
@@ -280,6 +281,11 @@ private void initVCFReader() {
                 + " MB.");
       }
 
+      if (bufferSizeMB > 2048) {
+        log.warn("Size of individfaual buffers is larger than 2048 MB. Consider setting a smaller " +
```
Shelnutt2: Put a note that this is a limitation of arrow 0.10.0 in spark 2.4, so people know. 

Also we need to take some action here. Should we error out or should we adjust the buffer now to 2048 max? What do you think the user would expect?

Also we should do the math for the user and tell them what is the max size they can set.

gsvic: I believe that we should adjust it to 2048. I also refactored the warning message.

PR #80, opened 2020-04-22
Add export test with non-contiguous samples

aaronwolen:
This adds a test for the segfault fixed in #79.


Shelnutt2:
@aaronwolen shall we merge this?

PR #81, opened 2020-04-24
Update to split coordinates for TileDB 2.0.0

Shelnutt2:
Instead of using the TILEDB_COORDS buffer now sample and end_pos are queried as independent buffers. This if a feature allowed in TileDB 2.0.0 which will improve performance and memory usage.

~We will wait to merge this until an official TileDB 2.0.0 release (which should be in a few days) and the TileDB superbuild verison will be updated then.~ TileDB 2.0.0 is released and this PR has been updated.

### DIFF
--- a/libtiledbvcf/src/dataset/attribute_buffer_set.cc
+++ b/libtiledbvcf/src/dataset/attribute_buffer_set.cc
@@ -25,43 +25,49 @@ void AttributeBufferSet::allocate_fixed(
   nbytes = mem_budget_mb == 0 ? 1024 : (mem_budget_mb * 1024 * 1024);
   num_offsets = nbytes / sizeof(uint64_t);
 
-  using names = TileDBVCFDataset::AttrNames;
+  using attrNames = TileDBVCFDataset::AttrNames;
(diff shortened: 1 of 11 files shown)


libtiledbvcf/src/dataset/attribute_buffer_set.cc
```diff
@@ -25,43 +25,49 @@ void AttributeBufferSet::allocate_fixed(
   nbytes = mem_budget_mb == 0 ? 1024 : (mem_budget_mb * 1024 * 1024);
   num_offsets = nbytes / sizeof(uint64_t);
 
-  using names = TileDBVCFDataset::AttrNames;
+  using attrNames = TileDBVCFDataset::AttrNames;
```
stavrospapadopoulos, line 28: `attr_names`, `dim_names` for consistency.

Shelnutt2: There is a parameter passed in to this function that is `attr_names` so we can't use that for the namespace alias. The alternative is to remove the alias and just reference the full namespace in the if statements below.

PR #86, opened 2020-05-05
Add CI job to build Dockerfiles

aaronwolen:
This updates the CI pipeline to build and push the docker images. It's setup so that images are only built when:

* `master` is updated, which will tag the image as `latest`, or
* a new `tag` is pushed, in which case the image is tagged with the tag version

Image building is implemented as a separate job, which depends on the *Build* job completing successfully. 

All 3 Dockerfiles are built in parallel, however, the python and dask images are commented out because they need to be updated.

Before merging this we need to 
- [x]  Add Docker Hub as registry service connection 
- [x] Create the corresponding repositories on Docker hub

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -30,7 +30,7 @@ jobs:
       architecture: 'x64'
   - template: ci/azure-linux_mac.yml
 
-- job:
+- job: Images
   dependsOn: Build
   condition: |
     or(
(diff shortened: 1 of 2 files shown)


ihnorton:
I'm not sure why we have three different images, but perhaps the `-cli` and `-py` ones could be combined.

aaronwolen:
Thanks, @ihnorton. I agree that it makes sense to consolidate the py/cli images.

Shelnutt2:
@aaronwolen the dockerhub registry service connection is named `dockerHub`

PR #89, opened 2020-05-06
Add test for ingesting variants with duped end pos

aaronwolen:
New CLI test for ingesting a VCF file containing variants with different start positions but share a common end position. Currently, this fails as expected with the `duplicate coordinates` error should pass once #82 is merged in.


aaronwolen:
Verified it passes with the fix [here](https://dev.azure.com/aaronwolen/aaron/_build/results?buildId=117&view=results).

aaronwolen:
Superseded by #90.

PR #90, opened 2020-05-07
Add tests for ingesting records with duplicate positions

aaronwolen:
This supersedes #89.

* As discussed, the CLI `--allow_duplicates` argument has been replaced with a `--no-duplicates` flag
* Adds tests for ingesting duplicates with and without the `--no-duplicates` flag


Shelnutt2:
Thanks for the changes and the unit test, this looks great!

PR #94, opened 2020-05-12
Fix cmake overrides for python setup

aaronwolen:
A couple small fixes:
- Setting `FORCE_EXTERNAL_HTSLIB=OFF` was ignored
- Several overrides were being passed to cmake in duplicate

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -106,17 +106,18 @@ def get_cmake_overrides():
     conf = list()
 
     key = "TILEDBVCF_CMAKE_PREFIX_PATH"
-    val = os.environ.get(key, None)
+    val = os.environ.get(key, default=None)
     if val:
         conf.append("-DCMAKE_PREFIX_PATH={}".format(val))
 
     key = "TILEDBVCF_FORCE_EXTERNAL_HTSLIB"
-    val = os.environ.get(key, None)
-    if val:
-        conf.append("-DFORCE_EXTERNAL_HTSLIB={}".format(val))
+    val = os.environ.get(key, default="ON")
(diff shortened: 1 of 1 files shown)


apis/python/setup.py
```diff
@@ -175,6 +172,7 @@ def find_or_build_libtiledbvcf(setuptools_cmd):
         p = PathConfig()
         package_data = []
         for obj in native_libs:
+            print("Copying file {0} to {1}".format(obj, p.pkg_src))
```
Shelnutt2, line 175: Left over debug or is the print on purpose?

apis/python/setup.py
```diff
@@ -128,15 +129,11 @@ def build_libtiledbvcf():
...
-                 '-DFORCE_EXTERNAL_HTSLIB=ON',
-                 '-DCMAKE_BUILD_TYPE=Release',
-                 src_dir]
+    cmake_cmd = [os.environ.get('CMAKE', 'cmake')]
+    cmake_cmd.extend(get_cmake_overrides())
+    cmake_cmd.append(src_dir)
 
-    env_conf = get_cmake_overrides()
-    cmake_cmd.extend(env_conf)
+    print("CMake configure command: {}".format(cmake_cmd))
```
Shelnutt2, line 136: Left over debug or is the print on purpose?

aaronwolen: Left over on purpose. I noticed we use them in [tiledb-py](https://github.com/TileDB-Inc/TileDB-Py/blob/d32b65667f8a10fce46be2e54ce5352178778385/setup.py#L205) and thought the feedback was helpful for options set via env variables.

No problem to take them out.

PR #99, opened 2020-05-18
Store VCFs by their start position instead of end position

joe-maley:
This change replaces the "end_pos" dimension with a "start_pos" dimension. The
"real_end_pos" attribute has been replaced with a "real_start_pos" attribute.

Notable changes:
1. We probably don't need the min-heap anymore because input is already sorted
in ascending order on start positions. To minimize this change set, the heap
is left as-is with two minor changes:
  a. Inputs are sorted on their start position (real or anchor).
  b. Nodes in the min-heap have a flag for determining if it the last node in
     a record. This is required because the ingestion algorithim currently
     expects a non-anchor node to signify the end of a record. With the
     non-anchor now at the front, the ingestion algorithm breaks.

2. The read path now queries the anchor gap to the left of the user's query
   instead of the right.

3. I had to re-create the unit test arrays ("ingestion_2samples",
   "ingestion_2samples_GT_DP_PL"). Two issues:
   a. In the exporter unit test, reader output may be re-arranged because the
      heap is no longer sorting by end point. The data is still correct, but
      in a different order.
   b. I don't know if this is a bug or user error: but I could not get the
      new arrays to contain the filter columns. I performed the same re-creation
      on `master` to verify this bug or user error is unrelated to this patch.
      I had to modify both the C++ reader unit test and the python unit test
      to no longer expect filters when reading.


joe-maley:
Closing this WIP PR.

PR #100, opened 2020-05-20
Store VCFs by their start position instead of end position

joe-maley:
This change replaces the "end_pos" dimension with a "start_pos" dimension. The
"real_end_pos" attribute has been replaced with a "real_start_pos" attribute.

This patch bumps the format version from V2 to V3. This patch is backwards
compatible for reading/exporting V2 arrays and ingesting new samples onto
an existing V2 array.

Design changes for backwards compatibility:
- The WriterWorker is now an abstract interface with two implementations:
  WriterWorkerV2 and WriterWorkerV3.
- RecordHeap has been renamed RecordHeapV2 and a new RecordHeapV3 has been
  introduced for the V3 write path.
- The read path has introduced a few branches and a v2/v3 function routines,
  rather than using entirely separate classes for each path.

Notable changes:
1. We probably don't need the min-heap anymore because input is already sorted
in ascending order on start positions. To minimize this change set, the V3 heap
is left as-is with two minor changes:
  a. Inputs are sorted on their start position (real or anchor).
  b. Nodes in the min-heap have a flag for determining if it the last node in
     a record. This is required because the ingestion algorithim currently
     expects a non-anchor node to signify the end of a record. With the
     non-anchor now at the front, the ingestion algorithm breaks.

2. The V3 read path now queries the anchor gap to the left of the user's query
   instead of the right.

3. I had to re-create the unit test arrays ("ingestion_2samples",
   "ingestion_2samples_GT_DP_PL"). Two issues:
   a. In the exporter unit test, reader output may be re-arranged because the
      heap is no longer sorting by end point. The data is still correct, but
      in a different order.
   b. I don't know if this is a bug or user error: but I could not get the
      new arrays to contain the filter columns. I performed the same re-creation
      on `master` to verify this bug or user error is unrelated to this patch.
      I had to modify both the C++ reader unit test and the python unit test
      to no longer expect filters when reading.

Note that the existing arrays have been moved from 'arrays' to 'arrays/v2' and
the new arrays are located in 'arrays/v3'.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -332,6 +332,7 @@ void Reader::read() {
 void Reader::init_for_reads() {
   read_state_.batch_idx = 0;
   read_state_.sample_batches = prepare_sample_batches();
+  read_state_.last_intersecting_region_idx_ = 0;
 
   init_exporter();
   prepare_regions(&read_state_.regions, &read_state_.query_regions);
@@ -577,8 +578,15 @@ bool Reader::process_query_results() {
         intersecting.second == std::numeric_limits<uint32_t>::max())
       continue;
 
-    // Report all intersections.
-    for (size_t j = intersecting.first; j <= intersecting.second; j++) {
+    // Report all intersections. If the previous read returned before
+    // reporting all intersecting regions, 'last_intersecting_region_idx_'
+    // will be non-zero. All regions with an index less-than
+    // 'last_intersecting_region_idx_' have already been reported, so we
+    // must avoid reporting them multiple times.
+    size_t j = read_state_.last_intersecting_region_idx_ > 0 ?
+                   read_state_.last_intersecting_region_idx_ :
+                   intersecting.first;
+    for (; j <= intersecting.second; j++) {
       const auto& reg = read_state_.regions[j];
       const uint32_t reg_min = reg.seq_offset + reg.min;
       const uint32_t reg_max = reg.seq_offset + reg.max;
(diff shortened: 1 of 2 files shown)


joe-maley:
@Shelnutt2 This patch contains the backwards compatibility logic from V3 to V2. The WIP branch that was previously up for review (here: https://github.com/TileDB-Inc/TileDB-VCF/pull/99) contains the logic for moving to the start position from end position. It may be easier to review that patch for the logic unrelated to backwards compatibility and then return to this patch to check the V2/V3 paths.

libtiledbvcf/test/src/unit-vcf-store.cc
```diff
@@ -448,4 +448,42 @@ TEST_CASE("TileDB-VCF: Test ingest 100", "[tiledbvcf][ingest]") {
 
   if (vfs.is_dir(dataset_uri))
     vfs.remove_dir(dataset_uri);
+}
+
+TEST_CASE("TileDB-VCF: Write to existing V2 array", "[tiledbvcf][ingest][v2]") {
```
joe-maley, line 453: This test modifies an existing array for purposes of ingesting new samples into an existing V2 array. It contains a shell-command to copy the array directory. I didn't want to spend the time introducing boost to do this in a safe cross-platform way because we're only running CI on Linux. If this was in the library code I would have at least used the POSIX apis, but left this as-is for a unit test.

joe-maley:
I've updated this PR with the recent fix (https://github.com/TileDB-Inc/TileDB-VCF/pull/104) in the V3 path. I've also tested this to work properly on the same V2 array that was tested on.

PR #102, opened 2020-05-30
Update TileDB to 2.0.3 for superbuild

Shelnutt2:
Update TileDB to 2.0.3 for superbuild


Shelnutt2:
> Nice. Should we bump to 0.4.2?

That is the plan. I've added the spark/java version bump to this PR. We'll include this and your README updates and do a 0.4.2 release.

PR #106, opened 2020-06-08
Fix over-allocating attribute buffers

joe-maley:
When VCF is given a memory budget of N, N/2 is given to the core while
the other N/2 are given to the VCF reader.

The VCF reader allocates M attribute buffers, where each buffer should have
a size of (N/2)/M, but it currently is allocated N/2 for _each_ buffer,
which over-allocates the VCF reader by a scale of M.

This fixes the issue.

### DIFF
--- a/libtiledbvcf/src/dataset/attribute_buffer_set.cc
+++ b/libtiledbvcf/src/dataset/attribute_buffer_set.cc
@@ -22,8 +22,10 @@ void AttributeBufferSet::allocate_fixed(
 
   // Requesting 0 MB will result in a 1 KB allocation. This is used by the
   // tests to test the path of incomplete TileDB queries.
-  nbytes = mem_budget_mb == 0 ? 1024 : (mem_budget_mb * 1024 * 1024);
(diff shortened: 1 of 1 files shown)


libtiledbvcf/src/dataset/attribute_buffer_set.cc
```diff
@@ -22,8 +22,10 @@ void AttributeBufferSet::allocate_fixed(
 
   // Requesting 0 MB will result in a 1 KB allocation. This is used by the
   // tests to test the path of incomplete TileDB queries.
-  nbytes = mem_budget_mb == 0 ? 1024 : (mem_budget_mb * 1024 * 1024);
-  num_offsets = nbytes / sizeof(uint64_t);
+  if (mem_budget_mb == 0) {
```
joe-maley, line 25: See line 20 for context. The issue is that the correct allocation is set on line 20, but we're overwriting it here (line 25 on the original) with the incorrect allocation by not dividing by num_buffers.

joe-maley:
@Shelnutt2 Is there any reason to hold this PR?

PR #110, opened 2020-06-18
Overhaul Dockerfile READMEs

aaronwolen:
New [README](https://github.com/TileDB-Inc/TileDB-VCF/blob/aw/ch2596/docker-quickstart/docker/README.md) for Dockerfiles that includes a quick start guide for `tiledbvcf-py` and `tiledbvcf-cli`. Mainly intended for Docker Hub. 

Other changes:

- final working directory is now `/data` rather than `/temp`
- `tiledbvcf-py` can be used to execute a script or launch an interactive session
- set `AWS_EC2_METADATA_DISABLED` env variable to avoid slow down caused by the (typically) necessary API calls

### DIFF
--- a/docker/Dockerfile-py
+++ b/docker/Dockerfile-py
@@ -56,4 +56,4 @@ RUN python setup.py install
 RUN rm -rf /tmp/tiledbvcf
 
 WORKDIR /data
-CMD ["python3"]
+ENTRYPOINT ["python3"]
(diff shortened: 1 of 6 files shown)


docker/Dockerfile-py
```diff
@@ -1,36 +1,8 @@
...
-# THE SOFTWARE.
-#
-# This Dockerfile builds the TileDB-VCF Python image.
-#
-
 FROM continuumio/miniconda3:4.8.2
 ENV CFLAGS "-march=haswell"
 ENV CPPFLAGS "-march=haswell"
 
+ENV AWS_EC2_METADATA_DISABLED true
```
Shelnutt2, line 5: Should we set this in all the dockerfiles?

aaronwolen: Probably makes sense for now but the issue isn't docker specific. I'm thinking we may need to add an option to core.

PR #118, opened 2020-07-07
VCF record iteration for overlapping records

joe-maley:
Currently, we buffer VCF records into memory and store the minimum start
position and maximum end position of these buffered records. When we iterate
to buffer the next set of VCF records, we seek to the maximum end position
and read from there. This breaks if there are any overlapping records.

This patch chanes the behavior from open+seek on each iteration to opening
once on the first iteration and making use of the internal htslib iterator
to move between records.

### DIFF
--- a/libtiledbvcf/src/utils/sample_utils.h
+++ b/libtiledbvcf/src/utils/sample_utils.h
@@ -47,15 +47,6 @@
 namespace tiledb {
 namespace vcf {
 
-/** Alias for unique_ptr to bcf_hdr_t. */
-typedef std::unique_ptr<bcf_hdr_t, decltype(&bcf_hdr_destroy)> SafeBCFHdr;
-
-/** Alias for unique_ptr to bcf1_t. */
-typedef std::unique_ptr<bcf1_t, decltype(&bcf_destroy)> SafeBCFRec;
-
-/** Alias for unique_ptr to htsFile. */
-typedef std::unique_ptr<htsFile, decltype(&hts_close)> SafeBCFFh;
-
 /** Struct holding information about available scratch disk space. */
 struct ScratchSpaceInfo {
   std::string path = "";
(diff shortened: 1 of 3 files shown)


joe-maley:
Taking this down for now. The CI found a flaky failure. I can repro locally.

joe-maley:
Re-opening, the CI failure reproduced on `master` so it is unrelated to this change.

For reference, this is the failure:
```
1: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1: tiledb_vcf_unit is a Catch v2.2.1 host application.
1: Run with -? for options
1:
1: -------------------------------------------------------------------------------
1: TileDB-VCF: Test export 100
1: -------------------------------------------------------------------------------
1: /Users/joe/Workspace/TileDB-VCF/libtiledbvcf/test/src/unit-vcf-export.cc:1239
1: ...............................................................................
1:
1: /Users/joe/Workspace/TileDB-VCF/libtiledbvcf/test/src/unit-vcf-export.cc:1239: FAILED:
1: due to unexpected exception with message:
1:   Error getting contig name from VCF; current record is null.
```

aaronwolen:
Thanks! I'm testing it with a large dataset that contains quite a few overlapping ranges.

joe-maley:
@aaronwolen This will solve the issue from last week where we increased the anchor size but still hit a failure. We will still see the same global ordering error message for some anchoring scenarios. Also note the newly discovered `Error getting contig name from VCF; current record is null.` error message.

aaronwolen:
Thanks, @joe-maley. I confirmed that increasing the anchor gap allows us to ingest samples containing overlaps but we still hit the expected global order error when using anchors narrower than the overlapping variants.

PR #119, opened 2020-07-07
 Add python support for scratch space

Shelnutt2:
Add python support for scratch space

Two new parameters are added to the ingest_samples for `scratch_space_path` and `scratch_space_size`. A c-api is also exposed.


Shelnutt2:
> Thanks, @Shelnutt2! I tested it out and hit an error because `set_scratch_space` wasn't added to `Writer`. I fixed it in another branch that also documents the new args. Should I make a PR against this branch?
> 
> Changes are [here](https://github.com/TileDB-Inc/TileDB-VCF/compare/sethshelnutt/ch2708/add-ability-to-set-scratch-space-from-python...aw/ch2708/add-ability-to-set-scratch-space-from-python?expand=1).

@aaronwolen I'm not sure how I missed that, must have missed commit the writer files. Please open a PR against this branch

aaronwolen:
NP, just added #122

PR #123, opened 2020-07-09
Heap-node ownership of buffered records

joe-maley:
This is a prerequisite to ingesting overlapping records.

Currently, the `VCF` class iterates through a VCF file with a readahead that
buffers multiple records into a single buffer. When the last record in the
buffer is read, the buffer is cleared and populated with the next set of
records. This is OK because we currently enforce that all records are ordered
and without overlap.

To support overlapping records, records may be processed out-of-order. We can
not safely clear the single buffer because nodes on the heap may still point
to them. This change moves ownership of the records from the `VCF` instance
to the individual nodes on the heap.

The `VCF` still performs a readahead, but places each record onto a queue. The
writer worker will pop the next record and place it into the heap. When the
queue is empty, the `VCF` instance will perform the next readahead. There is
no additional copying and the read IO is unchanged. As an optimization,
when the end node on a heap is popped, the allocated record is returned to the
`VCF` instance for re-use.

This patch is verbose because the `VCF` instance had to be split into a V2
and V3 variant to support backwards compatibility. Shared static utilities
have been placed into a `vcf/vcf_utils.h` file.

### DIFF
--- a/libtiledbvcf/src/vcf/vcf_utils.cc
+++ b/libtiledbvcf/src/vcf/vcf_utils.cc
@@ -1,0 +1,164 @@
+/**
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2020 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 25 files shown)


libtiledbvcf/src/vcf/vcf_utils.cc
```diff
@@ -0,0 +1,164 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "vcf/vcf_utils.h"
+
+namespace tiledb {
+namespace vcf {
+
+uint32_t VCFUtils::get_end_pos(
```
joe-maley, line 32: All routines here are a direct copy+paste from the existing `VCF`. There is no need to review the implementations.

libtiledbvcf/src/vcf/vcf_v3.cc
```diff
@@ -0,0 +1,397 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "vcf/vcf_v3.h"
+
+namespace tiledb {
+namespace vcf {
+
+VCFV3::VCFV3()
```
joe-maley, line 32: The VCFV3 is identical to the `VCF` with the following changes:
- `buffer_` is removed
- `buffer_offset_` is removed
- `record_queue_` is new
- `record_queue_pool_` is new
- `destroy_buffer` and `clear_buffer` have been removed
- `next` interface and implementation have been changed
- `return_record` is new
- `curr_record` is removed
- `read_records` implementation is changed

other misc stuff like `swap` and list initializer in the constructor
- `contig_name` accepts a record as an argument instead of using the `curr_record()`

PR #125, opened 2020-07-10
Support ingesting overlapping records

joe-maley:
This modifies the ingestion algorithm to support overlapping records. This
ensures anchor ranges are written in-order.


libtiledbvcf/src/vcf/vcf_v3.cc
```diff
@@ -127,20 +127,30 @@ bool VCFV3::is_open() const {
   return open_;
 }
 
-SafeSharedBCFRec VCFV3::next() {
+SafeSharedBCFRec VCFV3::peek_record() {
```
stavrospapadopoulos, line 130: Maybe `front` is a better name for this function?

PR #127, opened 2020-07-14
Add python method for retrieving attributes from a dataset

aaronwolen:
New `attributes()` method for `tiledbvcf.Dataset` that allows users to retrieve attributes available in a dataset. 

* by default all attributes are returned (with the exception of the aggregated `info` and `fmt` attributes)
* specific subsets of attributes can be selected with the `attr_type` argument

### DIFF
--- a/apis/python/src/tiledbvcf/binding/reader.cc
+++ b/apis/python/src/tiledbvcf/binding/reader.cc
@@ -353,6 +353,34 @@ std::string Reader::get_tiledb_stats() {
   return std::string(stats);
 }
 
+int32_t Reader::get_fmt_attribute_count() {
+  auto reader = ptr.get();
+  int32_t count;
+  check_error(reader, tiledb_vcf_reader_get_fmt_attribute_count(reader, &count));
+  return (count);
+}
+
+std::string Reader::get_fmt_attribute_name(int32_t index) {
+  auto reader = ptr.get();
+  char* name;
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbvcf/binding/reader.cc
```diff
@@ -349,6 +349,48 @@ std::string  Reader::get_tiledb_stats() {
   return std::string(stats);
 }
 
+int32_t Reader::get_fmt_attribute_count() {
```
Shelnutt2, line 352: It might be easier/better to combine these functions into a single one which returns the vector of names, something like:

```
std::vector<std::string> Reader::get_fmt_attributes() {
  auto reader = ptr.get();
  int32_t count;
  std::vector<std::string> ret;
  check_error(reader, tiledb_vcf_reader_get_fmt_attribute_count(reader, &count));

  for(int32_t index = 0; index < count; index++) {
    char* name;
    check_error(reader, tiledb_vcf_reader_get_fmt_attribute_name(reader, index, &name));
    ret.emplace_back(name);
  }

  return ret;

}
```

What do you think?

Your  python code becomes simpler then with something like:
```
def __fmt_attrs():
  return self.reader.get_fmt_attribute()
```

aaronwolen: Like it!

aaronwolen: Implemented this in 4496581—it definitely simplifies the python code. Thanks for the suggestion.

PR #136, opened 2020-07-28
Multi-stage build for CLI docker image

aaronwolen:
- updates the CLI Dockerfile to utilize multi-stage builds significantly shrinking the size of the final image (860Mb -> 187Mb)
- copies over `.git` so TileDB-VCF's version is now recorded at build time
- adds an example of creating an array with Docker, highlighting the need to specify user/group IDs to avoid permission issues

### DIFF
--- a/docker/Dockerfile-cli
+++ b/docker/Dockerfile-cli
@@ -4,7 +4,7 @@ ENV AWS_EC2_METADATA_DISABLED true
 
 # Install some dependencies
 RUN yum -y install which wget git tar gzip unzip gcc-c++ zlib-devel \
-       openssl-devel bzip2-devel tbb-devel libcurl-devel xz-devel make \
+       openssl-devel bzip2-devel libcurl-devel xz-devel make \
        automake autoconf \
     && yum clean all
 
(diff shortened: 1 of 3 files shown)


docker/Dockerfile-cli
```diff
@@ -16,14 +16,20 @@ RUN wget https://cmake.org/files/v3.12/cmake-3.12.3-Linux-x86_64.sh \
...
              -DOVERRIDE_INSTALL_PREFIX=OFF \
     && make -j4 \
     && make install-libtiledbvcf
-RUN rm -rf /tmp/libtiledbvcf
+
+FROM amazonlinux:latest
+COPY --from=builder /usr/local/bin/* /usr/local/bin/
+COPY --from=builder /usr/local/include/* /usr/local/include/
+COPY --from=builder /usr/local/lib/* /usr/local/lib/
+COPY --from=builder /usr/lib64/libtbb.so.2 /usr/lib64/libtbb.so.2
```
Shelnutt2, line 32: Is tbb needed?

Shelnutt2 (approved):
One minor comment, beside that lgtm

docker/Dockerfile-cli
aaronwolen: Nope. Just removed it—thanks.

PR #138, opened 2020-07-28
Add support for forcing TileDB to be built as part of superbuild

Shelnutt2:
This adds new support for forcing TileDB to built as part of the external build process. This works similar to HTSLIB.


aaronwolen:
Worked perfectly. 

Do you want to add something like 
```
if (FORCE_EXTERNAL_TILEDB)
  message(STATUS "Skipping search for TileDB, building it as an external project. To use system TileDB, set FORCE_EXTERNAL_TILEDB=OFF")
endif()
```

to be consistent with the `FORCE_EXTERNAL_HTSLIB` option?

Shelnutt2:
> Worked perfectly.
> 
> Do you want to add something like
> 
> ```
> if (FORCE_EXTERNAL_TILEDB)
>   message(STATUS "Skipping search for TileDB, building it as an external project. To use system TileDB, set FORCE_EXTERNAL_TILEDB=OFF")
> endif()
> ```
> 
> to be consistent with the `FORCE_EXTERNAL_HTSLIB` option?

Good point, I've added the message.

PR #140, opened 2020-07-30
Add unit tests which compares VCFInfoFmtDecoder [ch2819]

gsvic:


apis/java/src/test/java/io/tiledb/libvcfnative/VCFReaderTest.java
```diff
@@ -92,6 +93,122 @@ public void testReadCompletes() throws IOException {
...
+      reader.setBuffer(key, ByteBuffer.allocateDirect(1024).order(ByteOrder.nativeOrder()));
+    }
+
+    reader.setBufferOffsets(
+        "sample_name", ByteBuffer.allocateDirect(1024).order(ByteOrder.nativeOrder()));
+
+    reader.setBufferOffsets(
+        "info_END", ByteBuffer.allocateDirect(1024).order(ByteOrder.nativeOrder()));
+
+    reader.setBufferValidityBitmap(
```
Shelnutt2, line 135: Does any other field need validity buffers?

gsvic: I think that we should get advice from @aaronwolen for that, but for now we do not use it in the unit test even for the `info_END` attribute, this is a leftover from some test I've been doing during the development

PR #143, opened 2020-08-11
Add array and cli/py tests for querying records with null attributes

aaronwolen:
Adds cli/py tests for testing against bug fixed in #142

### DIFF
--- a/apis/python/tests/test_tiledbvcf.py
+++ b/apis/python/tests/test_tiledbvcf.py
@@ -352,7 +352,7 @@ def test_read_null_attrs(tmp_path):
 
     ds = tiledbvcf.Dataset(uri, mode='r')
     df = ds.read(attrs=['sample_name', 'pos_start', 'pos_end',
-                        'info_BaseQRankSum', 'info_DP', 'fmt_DP'],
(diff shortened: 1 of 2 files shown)


apis/python/tests/test_tiledbvcf.py
```diff
@@ -372,7 +372,9 @@ def test_read_null_attrs(tmp_path):
                                   np.array([89], dtype=np.int32), None,
                                   np.array([24], dtype=np.int32), None, None]),
             'fmt_DP': pd.Series([0, 0, 15, 64, 6, 2, 180,
-                                     88, 97, 24, 23, 21], dtype=np.int32)})
+                                     88, 97, 24, 23, 21], dtype=np.int32),
+            'fmt_MIN_DP': pd.Series([0, 0, 14, 30, 3, 1, 20,
+                                     None, 24, None, 23, 19], dtype=np.int32)})
```
Shelnutt2, line 377: I think you need to change the dtype to `"Int32"` or `pd.Int32Dtype()`, per https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html

aaronwolen: Updated the `dtype` as discussed and noted issue in ch2928.

PR #146, opened 2020-08-14
Add a flag for handling existing samples

Shelnutt2:
Allow a user to gracefully skip already registered sample in the dataset.

CLI flag is `--duplicate-sample`. Python option is `duplicate_sample` on `ingest_samples` function.


Shelnutt2:
@aaronwolen I believe we should wait to add support for handling the duplicate samples on store/ingestion. We need an efficient way to look up the sample id in the underlying tiledb array to check if it exists at all without a full query. We can't rely on the sample header because that might just be from registration.

gspowley:
This PR is a stale, since registration is no longer required. Ideas here will be captured in sc-23744.

PR #148, opened 2020-08-20
Added CI Stage for Building a Cross-Platform Jar

gsvic:
CI Stage that build a cross-platform JAR for
- Linux
- OSX


.azure-pipelines.yml
```diff
@@ -3,59 +3,159 @@ trigger:
...
+      steps:
+      - task: UsePythonVersion@0
+        inputs:
+          versionSpec: '$(python.version)'
+          architecture: 'x64'
+      - script: printenv
+        displayName: 'Print Env'
+      - template: ci/azure-linux_mac.yml
+
+    - job: Docker
```
Shelnutt2, line 40: @aaronwolen I think we should move the docker to a second stage instead of a conditional job, what are your thoughts? Its not directly related to this PR but since we are moving to stages, we could adjust this one too.

aaronwolen: Yes, I think that makes sense too. I'm about 1/2 done with stage-based refactor.

PR #153, opened 2020-08-27
Add limit_partitions kwarg option for read_dask/map_dask

ihnorton:
This option allows to stop reading after a given number of partitions (for debugging and resource estimation).

---

[ch3030]

### DIFF
--- a/apis/python/src/tiledbvcf/dask_functions.py
+++ b/apis/python/src/tiledbvcf/dask_functions.py
@@ -35,7 +35,7 @@ def _read_partition(read_args):
 
 
 def map_dask(self, fnc, attrs, region_partitions=1, sample_partitions=1,
-             samples=None, regions=None, samples_file=None, bed_file=None):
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbvcf/dask_functions.py
```diff
@@ -58,11 +57,17 @@ def map_dask(self, fnc, attrs, region_partitions=1, sample_partitions=1,
                                  samples_file, bed_file, fnc)
             partitions.append(dask.delayed(_read_partition)(read_args))
 
+            if (limit_partitions is not None and len(partitions) > limit_partitions):
+                break
+        else:
+            continue
+        break
```
ihnorton, line 64: FYI this pattern allows breaking out of the inner loop:

```
nops = 0
for i in range(10):
    for j in range(5):
        nops += 1
        if nops > 11:
            break
    else:
       continue
    break
print(i,j,nops) # -> 2 1 12
```

ihnorton: sneklang special

PR #160, opened 2020-09-02
Add timing for parsing and sorting bed file

Shelnutt2:
In verbose mode new output is given for bed file parsing and sorting times.


aaronwolen:
Great idea.

aaronwolen (approved):

PR #162, opened 2020-09-03
Add support for reading BED file in parallel

Shelnutt2:
If a bed file is passed and it has an tabix index we will parallelize the read and parsing based on the chromosomes in the index.


libtiledbvcf/src/vcf/region.cc
```diff
@@ -154,15 +155,74 @@ void Region::parse_bed_file_htslib(
...
     throw std::runtime_error("Error parsing BED file: " + bed_file_uri);
 
+  // If there is an index file process in parallel
+  if (regionsFile->tbx != nullptr) {
+    std::vector<std::future<std::list<Region>>> futures;
+    // Loop over the sequences in the index file and launch a task for each one
+    for (int i = 0; i < regionsFile->nseqs; i++) {
+      std::string chr = regionsFile->seq_names[i];
+      futures.emplace_back(
+          std::async(std::launch::async, [vfs, bed_file_uri, chr]() {
```
joe-maley, line 165: Capture `vfs`, `bed_file_uri`, and `chr` by reference to avoid a copy. This should be safe since we wait on the tasks to complete before these variables are destructed.

libtiledbvcf/src/vcf/region.cc
```diff
@@ -154,15 +155,74 @@ void Region::parse_bed_file_htslib(
...
+          regionsFile->end);
+    }
+  }
+
+  // reset htslib log level
+  hts_set_log_level(old_log_level);
+}
+
+std::list<Region> Region::parse_bed_file_htslib_section(
+    const VFS& vfs, const std::string& bed_file_uri, const std::string& chr) {
```
joe-maley, line 193: I don't see `vfs` used anywhere in this routine

aaronwolen (approved):

PR #163, opened 2020-09-07
V3 Arrays should move regions based on start_pos

Shelnutt2:
In v3 arrays the sort on the start instead of end means that we can't use the end position crossing to a new region to bump the minimum region checked for intersection. We need to instead wait for the start position to move to a new minimum region before shifting.

Limiting the regions we check for intersections is an optimization to limit the search domain as we move through the results of the query.


aaronwolen (approved):
LGTM!

PR #164, opened 2020-09-10
New C/Python APIs for retrieving dataset schema version and sample names

aaronwolen:
Adds the following python methods (as well as the requisite C APIs):
- `dataset.samples()`
- `dataset.schema_version()`

Retrieving the dataset's schema version may not be all that useful on its own but I plan on using it in a higher-level `describe()` function.

### DIFF
--- a/apis/python/src/tiledbvcf/binding/reader.cc
+++ b/apis/python/src/tiledbvcf/binding/reader.cc
@@ -356,6 +356,13 @@ int32_t Reader::get_dataset_version() {
   return version;
 }
 
+int32_t Reader::get_sample_count() {
+  auto reader = ptr.get();
+  int32_t count;
+  check_error(reader, tiledb_vcf_reader_get_sample_count(reader, &count));
+  return count;
+}
+
 std::vector<std::string> Reader::get_fmt_attributes() {
   auto reader = ptr.get();
   int32_t count;
--- a/apis/python/src/tiledbvcf/binding/reader.cc
+++ b/apis/python/src/tiledbvcf/binding/reader.cc
@@ -349,7 +349,7 @@ std::string Reader::get_tiledb_stats() {
   return std::string(stats);
 }
 
-int32_t Reader::get_dataset_version() {
+int32_t Reader::get_schema_version() {
   auto reader = ptr.get();
   int32_t version;
   check_error(reader, tiledb_vcf_reader_get_dataset_version(reader, &version));
--- a/apis/python/src/tiledbvcf/binding/reader.cc
+++ b/apis/python/src/tiledbvcf/binding/reader.cc
@@ -356,30 +356,6 @@ int32_t Reader::get_dataset_version() {
   return version;
 }
 
-int32_t Reader::get_sample_count() {
-  auto reader = ptr.get();
-  int32_t count;
-  check_error(reader, tiledb_vcf_reader_get_sample_count(reader, &count));
-  return count;
-}
-
-std::vector<std::string> Reader::get_sample_names() {
-  auto reader = ptr.get();
-  int32_t count;
-  std::vector<std::string> names;
-  check_error(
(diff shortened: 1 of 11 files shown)


libtiledbvcf/src/read/reader.cc
```diff
@@ -779,7 +779,7 @@ std::pair<size_t, size_t> Reader::get_intersecting_regions_v3(
   size_t original_last = last;
 
   // Next find the index of the last region that intersects the cell's
-  // REAL_START position. This is used as the actual interval of intersection.
+  // REAL_START position. This is used as the actual interval of intersection
```
joe-maley, line 782: Accidental? We should probably keep the period.

libtiledbvcf/src/read/reader.cc
```diff
@@ -1427,6 +1427,17 @@ void Reader::info_attribute_count(int32_t* count) {
   *count = this->dataset_->info_field_types().size();
 }
 
+void Reader::sample_count(int32_t* count) {
+  if (count == nullptr)
+    throw std::runtime_error(
+        "Error getting sample count; dataset is not open.");
```
joe-maley, line 1433: I don't think "dataset is not open" is the right description here.

libtiledbvcf/src/read/reader.cc
```diff
@@ -1427,6 +1427,17 @@ void Reader::info_attribute_count(int32_t* count) {
...
+  *count = dataset_->metadata().sample_names.size();
+}
+
+void Reader::sample_name(int32_t index, char** name) {
+  *name = const_cast<char*>(dataset_->metadata().sample_names[index].c_str());
```
joe-maley, line 1438: Rather than casting off the `const`, how about making the argument `const char** name`. This is probably the right thing to do anyway because we don't want the caller mutating it since they don't own it.

apis/python/src/tiledbvcf/binding/reader.cc
```diff
@@ -401,6 +408,28 @@ std::vector<std::string> Reader::get_queryable_attributes() {
...
+  check_error(reader, tiledb_vcf_reader_get_sample_count(reader, &count));
+  return count;
+}
+
+std::vector<std::string> Reader::get_sample_names() {
+  auto reader = ptr.get();
+  int32_t count;
+  std::vector<std::string> names;
+  check_error(reader, tiledb_vcf_reader_get_sample_count(reader, &count));
+
```
joe-maley, line 423: Add a `names.reserve(count);` here so that it pre-allocates exactly enough space for the names that we'll add. Otherwise this may realloc many times, possibly over-allocating memory.

joe-maley (approved):
LGTM, added a few minor comments you may want to address

libtiledbvcf/src/read/reader.cc
aaronwolen: Yeap, must have fat-fingered the delete key

aaronwolen:
Thanks, @joe-maley! I incorporated your suggestions.

libtiledbvcf/src/read/reader.cc
```diff
@@ -1427,6 +1426,16 @@ void Reader::info_attribute_count(int32_t* count) {
...
 }
 
+void Reader::sample_count(int32_t* count) {
+  if (count == nullptr)
+    throw std::runtime_error("Error getting sample count");
+  *count = dataset_->metadata().sample_names.size();
+}
+
+void Reader::sample_name(int32_t index, const char** name) {
+  *name = dataset_->metadata().sample_names[index].c_str();
```
Shelnutt2, line 1436: This will result in `name` having an invalid string reference. The `c_str()` is short lived and not guaranteed beyond the line it is called on.

Shelnutt2 (changes requested):

Shelnutt2 (approved):

PR #173, opened 2020-09-25
Validate END exists before updating on export

Shelnutt2:
Previously we always tried to update the END field in the info of the header for VCF record exports. This caused errors if the END field did not exist in the header. Now we conditionally update it.


aaronwolen (approved):
Perfect

aaronwolen:
Well... almost perfect 🙂

PR #179, opened 2020-09-29
Vn/remove max buffer elements

nguyenv:


ci/azure-linux_mac.yml.diff
```diff
@@ -0,0 +1,162 @@
+diff --git a/ci/azure-linux_mac.yml b/ci/azure-linux_mac.yml
```
Shelnutt2, line 1: Do we want this diff file?

nguyenv: Oh that was committed accidentally. I will remove it.

Shelnutt2 (approved):

PR #189, opened 2020-10-27
Validate region has same contig as record

Shelnutt2:
For v2/v3 arrays we need to validate region is same contig as record by checking the names. Even with global ordering in v2/v3 arrays its possible for a region to incorrectly overlap because a record extends before or after a contigs boundaries. Contig offsets are not guaranteed to be unique so we use a string comparison of the contig name.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -595,9 +595,9 @@ bool Reader::process_query_results_v3() {
     const uint32_t anchor_gap = dataset_->metadata().anchor_gap;
 
     // If we've passed into a new contig, get the new info for it.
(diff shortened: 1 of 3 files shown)


libtiledbvcf/src/read/reader.cc
```diff
@@ -624,6 +624,11 @@ bool Reader::process_query_results_v3() {
                    read_state_.region_idx;
     for (; j < read_state_.regions.size(); j++) {
       const auto& reg = read_state_.regions[j];
+
+      // If the region does not match the config skip
```
aaronwolen, line 628: Small typo: config -> contig

aaronwolen (approved):

PR #194, opened 2020-11-20
Switch to 3D Array

Shelnutt2:
This is a large PR which introduces a V4 ondisk format which is a 3D array. The dimensions are now `contig`, `start_pos` and `sample`. Contig and sample are now string fields and avoid the previous limitations of having to map unique sample names to integer ID fields.

We also switch the layout to row-major to preserve the same ordering with the new 3d array.


aaronwolen (approved):

libtiledbvcf/src/read/reader.cc
```diff
@@ -337,12 +337,21 @@ void Reader::read() {
 
 void Reader::init_for_reads() {
   read_state_.batch_idx = 0;
-  read_state_.sample_batches = prepare_sample_batches();
+  if (dataset_->metadata().version == TileDBVCFDataset::Version::V2 ||
+      dataset_->metadata().version == TileDBVCFDataset::Version::V3)
+    read_state_.sample_batches = prepare_sample_batches();
+  else
```
joe-maley, line 343: same here for preparing for a V5, maybe assert the `else` paths to `V4`.

libtiledbvcf/src/read/reader.cc
```diff
@@ -492,6 +572,19 @@ bool Reader::read_current_batch() {
     // Block on query completion.
     auto query_status = read_state_.async_query.get();
     read_state_.query_results.set_results(*dataset_, buffers_a.get(), *query);
+
+    // TODO: We should update all sizes here
```
joe-maley, line 576: Remove TODO or create a ticket?

libtiledbvcf/src/read/reader.cc
```diff
@@ -570,6 +674,147 @@ struct RegionComparator {
   }
 } RegionComparator;
 
+struct RegionComparatorV4 {
+  std::vector<Region>* regions_;
+
+  explicit RegionComparatorV4(std::vector<Region>* regions) {
+    regions_ = regions;
+  }
```
joe-maley, line 682: new-line after this bracket

libtiledbvcf/src/read/reader.cc
```diff
@@ -894,6 +1186,258 @@ std::vector<SampleAndId> Reader::prepare_sample_names() const {
...
+
+    r.seq_offset = contig_offset;
+    const uint32_t reg_min = r.min;
+    const uint32_t reg_max = r.max;
+
+    // Widen the query region by the anchor gap value, avoiding overflow.
+    uint64_t widened_reg_min = g > reg_min ? 0 : reg_min - g;
+    if (widened_reg_min <= regionNonEmptyDomain.second &&
+        reg_max >= regionNonEmptyDomain.first) {
+      filtered_regions.emplace_back(r);
```
joe-maley, line 1307: if `Region` has a move constructor, it looks like it makes sense to use that here since this is the last use of the element on `pre_partition_regions_list`:
```
filtered_regions.emplace_back(std::move(r));
```

libtiledbvcf/src/read/reader.cc
```diff
@@ -894,6 +1186,258 @@ std::vector<SampleAndId> Reader::prepare_sample_names() const {
...
+void Reader::prepare_regions_v4(
+    std::vector<Region>* regions,
+    std::unordered_map<std::string, std::vector<size_t>>*
+        regions_index_per_contig,
+    std::vector<std::pair<std::string, std::vector<QueryRegion>>>*
+        query_regions) const {
+  const uint32_t g = dataset_->metadata().anchor_gap;
+  // Use a linked list for pre-partition regions to allow for parallel parsing
+  // of BED file
+  std::list<Region> pre_partition_regions_list;
```
joe-maley, line 1255: I'm not sure what a `list` buys us over a `vector` here. Note that if this contains a large number of elements, the destruction time can be significantly larger because it needs to deallocate each individual element rather than deallocating a single virtually-contiguous buffer in a vector.

libtiledbvcf/src/utils/buffer.h
```diff
@@ -110,6 +131,10 @@ class Buffer {
 
   uint64_t data_size_;
 
+  uint64_t data_effective_size_;
+
+  uint64_t offset_nelts_;
```
joe-maley, line 136: add these to the default constructor

libtiledbvcf/src/utils/sample_utils.cc
```diff
@@ -170,5 +170,30 @@ std::vector<std::vector<SampleAndIndex>> batch_elements_by_tile(
...
+  for (unsigned vec_idx = 0; vec_idx < vec.size(); vec_idx++) {
+    auto sample = vec[vec_idx];
+    // When batching we must include samples only in the same tile extent
+    if (count >= tile_size) {
+      // reset count
+      count = 0;
+      result.emplace_back(batch);
+      batch = std::vector<SampleAndIndex>();
+    }
+    batch.push_back(vec[vec_idx]);
```
joe-maley, line 188: emplace_back, also on L193

libtiledbvcf/src/utils/sample_utils.h
```diff
@@ -238,6 +239,18 @@ class SampleUtils {
 std::vector<std::vector<SampleAndIndex>> batch_elements_by_tile(
     const std::vector<SampleAndIndex>& vec, uint64_t tile_size);
 
+/**
+ * Batches the given vector into a vector of vectors based on the tile_size.
+ * Ideally the vectors will be fixed (even) sized. If there is not an evenly
+ * divisible number of samples the last batch will be smaller
```
joe-maley, line 245: period at end of sentence

libtiledbvcf/src/vcf/region.h
```diff
@@ -96,6 +96,14 @@ struct Region {
       const std::map<std::string, uint32_t>& contig_offsets,
       std::vector<Region>* regions);
 
+  /**
+   * Sorts the given list of Regions by their global offsets, using the provided
+   * offsets map.
+   *
+   * @param regions Vector of regions to be sorted (in-place)
+   */
+  //  static void sort_v4(std::vector<Region>* regions);
```
joe-maley, line 105: remove

libtiledbvcf/src/vcf/vcf_v4.cc
```diff
@@ -0,0 +1,404 @@
+/**
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2017-2018 TileDB, Inc.
```
joe-maley, line 6: 2020

libtiledbvcf/src/vcf/vcf_v4.h
```diff
@@ -0,0 +1,200 @@
+/**
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2017-2018 TileDB, Inc.
```
joe-maley, line 6: 2020

libtiledbvcf/src/vcf/vcf_v4.h
```diff
@@ -0,0 +1,200 @@
...
+        uint32_t pos);
+    void reset();
+    void swap(Iter& other);
+    bool next(bcf1_t* rec);
+
+   private:
+    SafeBCFFh fh_;
+    bcf_hdr_t* hdr_;
+    hts_itr_t* hts_iter_;
+    tbx_t* tbx_;
```
joe-maley, line 156: assign these to zero either in-line or in the default constructor

libtiledbvcf/src/vcf/vcf_v4.cc
```diff
@@ -0,0 +1,404 @@
...
+
+#include "vcf/vcf_v4.h"
+
+namespace tiledb {
+namespace vcf {
+
+VCFV4::VCFV4()
+    : open_(false)
+    , path_("")
+    , index_path_("")
```
joe-maley, line 35: not needed, we can let these default-construct by leaving them out of this initi-list

libtiledbvcf/src/vcf/vcf_v4.h
```diff
@@ -0,0 +1,200 @@
...
+   */
+  void pop_record();
+
+  /**
+   * As an optimization, records returned from `next()` may be passed
+   * to this routine for re-use to reduce memory allocation overhead.
+   *
+   * @param record The processed record to return into the allocation pool.
+   */
+  void return_record(SafeSharedBCFRec record);
```
joe-maley, line 119: I don't see this used anywhere, but it could be `SafeSharedBCFRec&&` to avoid a copy if the caller is done with the `record`.

libtiledbvcf/src/write/record_heap_v4.cc
```diff
@@ -0,0 +1,84 @@
+/**
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2019 TileDB, Inc.
```
joe-maley, line 6: 2020

libtiledbvcf/src/write/record_heap_v4.h
```diff
@@ -0,0 +1,110 @@
+/**
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2019 TileDB, Inc.
```
joe-maley, line 6: 2020

libtiledbvcf/src/write/record_heap_v4.h
```diff
@@ -0,0 +1,110 @@
...
+
+class RecordHeapV4 {
+ public:
+  enum class NodeType { Record, Anchor };
+  struct Node {
+    Node()
+        : vcf(nullptr)
+        , type(NodeType::Record)
+        , record(nullptr)
+        , contig()
```
joe-maley, line 46: not needed, this is implicit

libtiledbvcf/src/write/writer_worker_v4.cc
```diff
@@ -0,0 +1,408 @@
+/**
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2019 TileDB, Inc.
```
joe-maley, line 6: 2020

libtiledbvcf/src/write/writer_worker_v4.cc
```diff
@@ -0,0 +1,408 @@
...
+ * THE SOFTWARE.
+ */
+
+#include "write/writer_worker_v4.h"
+
+namespace tiledb {
+namespace vcf {
+
+WriterWorkerV4::WriterWorkerV4()
+    : dataset_(nullptr) {
```
joe-maley, line 33: let's init records_buffered_ and anchors_buffered_ here as well

joe-maley (commented):
misc things below:

libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -377,11 +397,68 @@ void TileDBVCFDataset::load_field_type_maps(const tiledb::Context& ctx) {
...
+  std::vector<SampleAndIndex> samples = SampleUtils::build_samples_uri_list(
+      vfs, params.sample_uris_file, params.sample_uris);
+  if (samples.empty())
+    throw std::invalid_argument(
+        "Cannot register samples; samples list is empty.");
+
+  // Registration proceeds in batches, double-buffering the downloads.
+  std::set<std::string> sample_set(
+      metadata_.all_samples.begin(), metadata_.all_samples.end());
+  std::map<std::string, std::string> sample_headers;
```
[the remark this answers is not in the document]
Shelnutt2: There is no need for them to be ordered. I don't think we have any maps in anywhere that are needed. Any sorting we do is on vectors.

libtiledbvcf/src/read/reader.cc
Shelnutt2: The reason we use a list is in the `Region::parse_bed_file_htslib` we parallelize the parsing on a per contig bases. We then `splice` the lists together without copying. We eventually sort/filter the regions below and produce a vector which we carry through the entire read.

libtiledbvcf/src/vcf/vcf_v4.h
Shelnutt2: It is used here: https://github.com/TileDB-Inc/TileDB-VCF/blob/dd3b459f4f682f6c02c9ed8e0ffab87c84d2f822/libtiledbvcf/src/write/writer_worker_v4.cc#L155

libtiledbvcf/src/dataset/tiledbvcfdataset.cc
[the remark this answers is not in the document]
Shelnutt2: @joe-maley I opened a separate story for this. There is a lot of map usage and in the new v4 system not much of it needs to be ordered. I feel its better to do this in a follow up PR because many basic functions and class attributes needs to be adjusted if we do it across the board, which I think we should.

PR #198, opened 2020-12-04
Build Chunked Arrow Table in _read_partition()

nguyenv:


apis/python/src/tiledbvcf/dask_functions.py
```diff
@@ -24,13 +26,21 @@ def _read_partition(read_args):
     df = ds.read(attrs=read_args.attrs, samples=read_args.samples,
                  regions=read_args.regions, samples_file=read_args.samples_file,
                  bed_file=read_args.bed_file)
+
+    table_fragments = [pa.Table.from_pandas(df)]
+    while not ds.read_completed():
+        ds.reader.read()
+        pa_fragment = ds.reader.get_results_arrow()
+        pydict_fragment = pa_fragment.to_pydict()
+        table_fragments.append(pa.Table.from_pydict(
```
Shelnutt2, line 35: Can we avoid the pydictionary to and from and use the output from `get_results_arrow`? `get_results_arrow` returns a array table object?

nguyenv: Yes let me change that. I forgot the only reason I was trying to do this conversions before was to get around the memory leak issue.

nguyenv: I take that back. If remove the intermediary step of converting to a pydict, then it fails the `test_dask.py` unit tests. I will message you in Slack because I can't seem to correctly format my code or attach text files here.

Shelnutt2 (approved):

ihnorton (approved):

PR #200, opened 2020-12-04
Add python options to assist dask dataframe

Shelnutt2:
Two new options were added to read/continue_read to specify if we should return a pandas dataframe or the arrow table. Also added was an option for if the previous results should be freed before continuing the read.

This will allow for the dask dataframe to more efficiently build from the arrow table.


nguyenv (approved):
Thanks for the `continue_read()` mod.

Shelnutt2:
Incorporated into #198

PR #201, opened 2020-12-04
Use prebuilt artifacts by default for TileDB

Shelnutt2:
This speeds up the build time significantly by using prebuilt artifacts for TileDB if its not on the build system. A new option `DOWNLOAD_TILEDB_PREBUILT` is added to allow the user to toggle building from source or using the prebuilt artifacts.


Shelnutt2:
@joe-maley @normanb @ihnorton @gsvic this PR allows the `TileDB-VCF` cmake superbuild to use TileDB prebuilt artifacts, instead of always building TileDB from source. I thought this might be of general interest since we end up using a `FindTileDB_EP` style superbuild in a lot of projects.

gsvic (approved):
Looks great! LGTM

aaronwolen (approved):

PR #208, opened 2020-12-10
Sort v4 regions in lexicographical order for writes

Shelnutt2:
Sort v4 regions in lexicographical order for writes

### DIFF
--- a/libtiledbvcf/src/vcf/region.h
+++ b/libtiledbvcf/src/vcf/region.h
@@ -62,6 +62,17 @@ struct Region {
   std::string to_str(Type type = Type::ZeroIndexedInclusive) const;
 
   /**
+   * Comparator used in sorting by contig name
+   * @param other
+   * @return
+   */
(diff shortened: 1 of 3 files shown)


aaronwolen (approved):

libtiledbvcf/src/vcf/region.h
```diff
@@ -61,6 +61,18 @@ struct Region {
...
+  /**
+   * Comparator used in sorting by contig name
+   * @param other
+   * @return
+   */
+  bool operator<(const Region& other) const {
+    return (seq_name < other.seq_name);
+    if (seq_name == other.seq_name)
+      return min < other.min;
+    return seq_name < other.seq_name;
```
joe-maley, line 73: This is dead code, is L70 a mistake?

Shelnutt2: Thanks, fixed

joe-maley (approved):

PR #214, opened 2020-12-14
Ingestion improvements

Shelnutt2:
This PR adds 3 improvements to ingestion:

1) A new memory budget parameter for setting the max size of the TileDB query buffer for write
2) Store the contig used for seek on the VCF record
3) Add more verbose output on ingestion to show the write counts

### DIFF
--- a/libtiledbvcf/src/vcf/vcf_v4.h
+++ b/libtiledbvcf/src/vcf/vcf_v4.h
@@ -192,6 +192,9 @@ class VCFV4 {
 
   /** Swap all fields with the given VCFV3 instance. */
   void swap(VCFV4& other);
+
+  /** name of contig last seeked to. */
+  std::string seeked_contig_name_;
 };
 
 }  // namespace vcf
--- a/libtiledbvcf/src/write/writer.cc
+++ b/libtiledbvcf/src/write/writer.cc
@@ -429,6 +429,12 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples(
             throw std::runtime_error(
                 "Error submitting TileDB write query; unexpected query "
                 "status.");
+
+          if (ingestion_params_.verbose)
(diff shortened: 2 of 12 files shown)


libtiledbvcf/src/vcf/vcf_v4.h
```diff
@@ -192,6 +192,9 @@ class VCFV4 {
 
   /** Swap all fields with the given VCFV3 instance. */
   void swap(VCFV4& other);
+
+  /** name of contig last seeked to. */
+  std::string contig_name_;
```
aaronwolen, line 197: Could this be `last_contig_name` or `prev_contig_name` for clarity?

libtiledbvcf/src/write/writer.cc
```diff
@@ -429,6 +431,12 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples(
             throw std::runtime_error(
                 "Error submitting TileDB write query; unexpected query "
                 "status.");
+
+          if (ingestion_params_.verbose)
```
aaronwolen, line 435: Definitely a fan of more info for the user during ingestion but this will be printing quite a bit of text. Any change it could negatively impact performance?

libtiledbvcf/src/write/writer.cc
```diff
@@ -429,6 +431,12 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples(
...
+            std::cout << "Writing " << worker->records_buffered()
+                      << " for contig " << worker->region().seq_name
+                      << " (task " << i << " / " << tasks.size() << ")"
+                      << std::endl;
```
aaronwolen, line 439: How about something like "Recorded `x` cells for contig `y`" to be more inline with our export verbose messages (e.g., Processed 79133 cells in 0.153361 sec.). 

Also `i +1` for tasks?

PR #215, opened 2020-12-14
Don't batch v4 reads by samples

Shelnutt2:
Users can still partition the read but we don't sub-divide a partition by the sample batch count (tile extent). This is not longer needed nor optimal in v4 arrays because sample is now the last dimension.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -1197,20 +1197,7 @@ std::vector<std::vector<SampleAndId>> Reader::prepare_sample_batches_v4()
       params_.sample_partitioning.num_partitions,
       &samples);
 
-  // Group partition into space tile batches.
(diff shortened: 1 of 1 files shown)


joe-maley (approved):
LGTM but I don't remember the motivation for this. Could you remind me?

Shelnutt2:
> LGTM but I don't remember the motivation for this. Could you remind me?

The original motivation for such a feature in v2/v3 code is that we would run a query which more closely aligned to the tile extent size of the uint32_t sample dimension. This would allow for more performant queries.

The motivation for removing this ins v4, is 1) There is no longer tile extent, we still write in batches of 10 (configurable) samples but the sample is now the 3rd dimension instead of the 1st dimension. 2) There is no way for us to guarantee the samples being queried in the batch of 10 actually line with the "ingestion batch of 10", so we might not really have the effect we are looking for here. Since the read will be on effectively random samples, group by 10 doesn't force any actual alignment.

PR #221, opened 2020-12-15
Handle all sample export for v4 with optimal range

Shelnutt2:
In v4 since we have strings core will not coalesce the point ranges. In the special case that we are exporting all samples, we can optimize by only setting one range with the first and last sample of the partition/batch.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -379,7 +379,8 @@ void Reader::init_for_reads_v3() {
 void Reader::init_for_reads_v4() {
   assert(dataset_->metadata().version == TileDBVCFDataset::Version::V4);
   read_state_.batch_idx = 0;
-  read_state_.sample_batches = prepare_sample_batches_v4();
+  read_state_.sample_batches =
+      prepare_sample_batches_v4(&read_state_.all_samples);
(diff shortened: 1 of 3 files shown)


libtiledbvcf/src/read/reader.h
```diff
@@ -495,16 +499,24 @@ class Reader {
 
   /**
    * Prepares sample batches for export
-   * @return
+   * @param all_samples Boolean to indicated if all samples are being queried by
```
joe-maley, line 502: I would have the word "output" somewhere in this description , same for the L515.

libtiledbvcf/src/read/reader.cc
```diff
@@ -1171,10 +1186,10 @@ std::vector<std::vector<SampleAndId>> Reader::prepare_sample_batches() const {
   return result;
 }
 
-std::vector<std::vector<SampleAndId>> Reader::prepare_sample_batches_v4()
-    const {
+std::vector<std::vector<SampleAndId>> Reader::prepare_sample_batches_v4(
+    bool* all_samples) const {
```
joe-maley, line 1190: I would `assert(all_samples)` here and in the other routine.

joe-maley (approved):
LGTM, minor nits

aaronwolen (approved):

PR #227, opened 2020-12-22
Parallelize finalize calls for v4

Shelnutt2:
The finalize calls on all the fragments can add up to detrimental performance because it blocks further writes. This change launches all finalizes in separate non-blocking threads.


Shelnutt2:
Tagging @joe-maley and @aaronwolen just for awareness of the change.

stavrospapadopoulos (approved):

PR #231, opened 2020-12-27
Add new `--sample-batch-size` for store command

Shelnutt2:
This allow for setting the number of samples to batch per ingestion. This removes the parameter from the create of the dataset as its no longer required to be uniform for all ingestions.

This also adds the equivalent c-api and support in python.

### DIFF
--- a/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
+++ b/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
@@ -129,7 +129,6 @@ void TileDBVCFDataset::create(const CreationParams& params) {
   create_group(ctx, params.uri);
 
   Metadata metadata;
-  metadata.row_tile_extent = params.row_tile_extent;
   metadata.tile_capacity = params.tile_capacity;
   metadata.anchor_gap = params.anchor_gap;
   metadata.extra_attributes = params.extra_attributes;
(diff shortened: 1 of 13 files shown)


libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -521,7 +520,8 @@ void TileDBVCFDataset::print_dataset_stats() {
   utils::enable_pretty_print_numbers(std::cout);
   std::cout << "Statistics for dataset '" << root_uri_ << "':" << std::endl;
   std::cout << "- Version: " << metadata_.version << std::endl;
-  std::cout << "- Row tile extent: " << metadata_.row_tile_extent << std::endl;
+  std::cout << "- Row tile extent: " << metadata_.ingestion_sample_batch_size
```
aaronwolen, line 523: Should we change the stat field name too?

libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -1117,7 +1119,8 @@ void TileDBVCFDataset::read_metadata() {
     return read_metadata_v4();
 
   get_md_value("tile_capacity", TILEDB_UINT64, &metadata.tile_capacity);
-  get_md_value("row_tile_extent", TILEDB_UINT32, &metadata.row_tile_extent);
+  get_md_value(
+      "row_tile_extent", TILEDB_UINT32, &metadata.ingestion_sample_batch_size);
```
aaronwolen, line 1123: Shouldn't the metadata key be `ingestion_sample_batch_size`?

aaronwolen (approved):

PR #243, opened 2021-01-07
C/Python APIs for writer config params

aaronwolen:
Extends the C API to allow setting the following dataset creation/ingestion parameters and exposes them to Python:

**Creation parameters**
- `anchor_gap`
- `tile_capacity`

**Ingestion parameters**
- `num_threads`
- `thread_task_size`
- `memory_budget_mb`
- `max_buffer_limit`

### DIFF
--- a/apis/python/src/tiledbvcf/dataset.py
+++ b/apis/python/src/tiledbvcf/dataset.py
@@ -211,8 +211,9 @@ class Dataset(object):
         self,
         extra_attrs=None,
         tile_capacity=None,
+        anchor_gap=None,
         checksum_type=None,
-        allow_duplicates=True
+        allow_duplicates=True,
     ):
         """Create a new dataset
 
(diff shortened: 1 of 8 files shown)


apis/python/src/tiledbvcf/dataset.py
```diff
@@ -258,6 +276,9 @@ def ingest_samples(
         if sample_uris is None:
             return
 
+        if memory_budget is None:
```
Shelnutt2, line 279: `is not None` ?

libtiledbvcf/src/c_api/tiledbvcf.cc
```diff
@@ -870,6 +893,13 @@ int32_t tiledb_vcf_writer_get_last_error(
   return TILEDB_VCF_OK;
 }
 
+int32_t tiledb_vcf_writer_set_memory_budget(
+    tiledb_vcf_writer_t* writer, uint64_t size) {
+  writer->writer_->set_memory_budget(size);
```
Shelnutt2, line 898: Do you need the sanity checks and save_error_catch?

aaronwolen: Right. Shall I add it to `tiledb_vcf_writer_set_scratch_space` right below?

Shelnutt2: Yes good catch on that one missing it too.

PR #246, opened 2021-01-14
C/Python APIs for vacuuming/consolidating fragments and fragment metadata

aaronwolen:


aaronwolen:
Closing this so we can implement as a separate class.

PR #247, opened 2021-01-21
Add verbose output and logging of buffer sizes

Shelnutt2:
This helps to understand the size of the core C++ buffer sizes or the spark buffer sizes.


gsvic (approved):
Looks good!

PR #259, opened 2021-02-05
Forward TileDB_DIR to the TileDB EP so it is used

ihnorton:


libtiledbvcf/cmake/Superbuild.cmake
```diff
@@ -53,6 +53,7 @@ set(INHERITED_CMAKE_ARGS
   -DFORCE_EXTERNAL_HTSLIB=${FORCE_EXTERNAL_HTSLIB}
   -DFORCE_EXTERNAL_TILEDB=${FORCE_EXTERNAL_TILEDB}
   -DTILEDB_S3=${TILEDB_S3}
+  -DTileDB_DIR=${TileDB_DIR}
```
Shelnutt2, line 56: Should this be `TILEDB_DIR` (all caps)?

ihnorton: That doesn't work on case-sensitive file systems because core ships `TileDBConfig.cmake`. When calling `find_package(TILEDB)` cmake looks for `TileDBConfig.cmake` and `tiledb-config.cmake`.

Shelnutt2: Makes perfect sense, thanks for clarifying. The case just caught my eye but I missed that its for the `find_package` to match the config.cmake file.

Shelnutt2 (approved):

PR #265, opened 2021-02-09
Check contig list for v4 before adding regions

Shelnutt2:
As an optimization to avoid useless regions, we now optionally validate the contig is present in the list of contigs from the VCF header data. This is an optimization to avoid running queries which are inside the non-empty domain but don't actually exist.


aaronwolen (approved):

Shelnutt2:
This conflicts with the changes to minimally fetch the headers, closing.

PR #270, opened 2021-02-13
Add % of query processed in verbose output

Shelnutt2:
This is just an estimate based on the TileDB Query estimated result size. However it provides a decent guess to the user about how much data is left to query.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -25,6 +25,9 @@
  */
 
 #include <future>
+#include <iomanip>
+#include <random>
+#include <thread>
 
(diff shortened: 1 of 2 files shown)


aaronwolen (approved):

Shelnutt2:
If there is a large number of ranges this can cause a delay in starting the query as computing the estimates can be costly. However  I believe we can make an adjustment in core for this. Computing the estimates will always be costly, however the estimate is computed when `query->submit()` is called anyway. If we cache the results in core then the cost of this new call will go away as no matter what we have to compute it inside the `submit`.

For now I've adjusted this branch to wrap the compute only in verbose mode.

PR #277, opened 2021-02-15
Add detection for all samples in user passed list

Shelnutt2:
This adds support for the optimization path where the user passes a list of samples, but it actually the entire array export. Previously we handled this only if the user didn't pass a list.


Shelnutt2:
This is failing because of an ordering problem in our incomplete query tests. With the optimizations the results are coming back in a different order in the incomplete queries.

gspowley:
Closing stale PR.

PR #286, opened 2021-02-28
 Avoid unnecessary fetching of sample list for v4

Shelnutt2:
If the user is exporting all samples, we can avoid actually getting the full list of samples names in most cases. This removes a query to the vcf_header array.

Depends on #285


libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -646,10 +646,27 @@ std::shared_ptr<tiledb::Array> TileDBVCFDataset::open_data_array(
...
+        "request");
+
+  if (all_samples && !samples.empty())
+    throw std::runtime_error(
+        "Cannot set all_samples and samples list in same fetch vcf headers "
+        "request");
+
+  if (first_sample && !samples.empty())
+    throw std::runtime_error(
+        "Cannot set first_sample and samples lsit in same fetch vcf headers "
```
aaronwolen, line 667: Typo: lsit -> list

aaronwolen (approved):

PR #313, opened 2021-04-28
Update htslib to version 1.12

aaronwolen:


Shelnutt2:
@aaronwolen you'll also need to update the java api [native loaded for htslib](https://github.com/TileDB-Inc/TileDB-VCF/blob/813d4226c5407bbd40f5f6ab36fc264fb628e196/apis/java/src/main/java/io/tiledb/libvcfnative/NativeLibLoader.java#L98)

gspowley:
Release 0.14.0 uses htslib 1.15.

PR #314, opened 2021-04-28
Patch htslib 1.10 to fix build with autoconf 2.70+

ihnorton:
from and details in: https://github.com/samtools/htslib/pull/1198


ihnorton:
@aaronwolen we will likely need to update the feedstock to require autotools after this change (if it is not installed by default).

Shelnutt2:
@ihnorton I think we need to add `automake` from brew now too.

Shelnutt2 (approved):

PR #330, opened 2021-05-09
Support for merging contigs into single fragments

Shelnutt2:
This is a (large) performance optimization in cases where the samples have a large number of pseudo configs which are small. These small contigs create a large number of fragments which can cause slower listing of the array on S3/GCS/Azure Blob Store object stores. This makes opening the array take significantly longer in certain scenarios. The change here allows the user to set two list (keep separate and mergeable) of contigs which are allowed to be merged into a single fragment in the sample batch.

The default `keep-as-separate` list will keep human contigs in USCS or ensembl format from being merged. For performance of reads, it is desirable to keep contigs which are going to read frequently or are large in size separate and not merged.

Further merge takes into account (automatically) lexicographical ordering. That is in the event of chromsomes `chr1`, `chr1_x`, `chr1_y`, `chr2`, and `chr2_x`, we will merge `chr1_x` and `chr1_y` together but not merge `chr2_x` with `chr1_y` because we must avoid overlapping the fragment's non empty domain with `chr2`.

Depends on #316  #329  #328

### DIFF
--- a/libtiledbvcf/src/write/writer.cc
+++ b/libtiledbvcf/src/write/writer.cc
@@ -549,8 +549,9 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples_v4(
         }
       }
     } catch (const std::exception& e) {
-      //      std::cout << "sample batch [" << first_sample_name << ", " <<
-      //      last_sample_name << "] not found in existing array" << std::endl;
+      //      std::cout << "sample batch [" << first_sample_name << ", "
+      //                << last_sample_name << "] not found in existing array"
+      //                << std::endl;
     }
   }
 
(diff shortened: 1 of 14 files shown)


libtiledbvcf/src/write/writer.cc
```diff
@@ -655,22 +752,59 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples_v4(
...
+
+              // Set new contig
+              last_region_contig = contig;
+              starting_region_contig_for_merge = contig;
+            } else {
+              if (ingestion_params_.verbose)
+                std::cout
+                    << "last contig " << last_contig_mergeable << " and "
+                    << "contig " << contig
+                    << " found to be mergeable, combing into super fragment"
```
aaronwolen, line 803: s/combing/combining/

aaronwolen (approved):

PR #335, opened 2021-06-14
Add spark option to setting specific partition idx

Shelnutt2:
This allows the user to set a single partition or partition index for ranges/samples like is supported in other APIs. This is helpful when debugging.


ihnorton (approved):
LGTM but probably want ack from Java eyes as well :)

DimitrisStaratzis (approved):

PR #336, opened 2021-06-14
Add debugging options for printing query details

Shelnutt2:
This adds three new reader options for printing the read details including the sample list used, the vcf regions passed and the actual TileDB query ranges performed.

Three new CLI options are added:
```
    --debug-print-vcf-regions
                         Enable debug printing of vcf region passed by user or bed file. Requires verbose mode

    --debug-print-sample-list
                         Enable debug printing of sample list used in read. Requires verbose mode

    --debug-print-tiledb-query-ranges
                         Enable debug printing of tiledb query ranges used in read. Requires verbose mode
```

Three spark options where added:

`debug.print_vcf_regions`
`debug.print_sample_list`
`debug.print_tiledb_query_ranges`

Python is implemented in pybind11 but the user interface is not defined yet. This will be a followup with change to how we can more easily pass all the various debug options we want to set.

### DIFF
--- a/apis/java/src/main/java/io/tiledb/libvcfnative/LibVCFNative.c
+++ b/apis/java/src/main/java/io/tiledb/libvcfnative/LibVCFNative.c
@@ -805,3 +805,42 @@ JNIEXPORT jint JNICALL Java_io_tiledb_libvcfnative_LibVCFNative_tiledb_1vcf_1rea
 
     return rc;
 }
+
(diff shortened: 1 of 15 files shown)


Shelnutt2:
@aaronwolen I'm trying to decide about the output format. Should we print a list of values or use json? Can you try out the new CLI options on this branch when you have time and tell me which way you like better?

aaronwolen (approved):

PR #342, opened 2021-06-28
Remove double buffering

Shelnutt2:
The double buffering is not worth it. The processing time in TileDB-VCF is always a fraction of the time spent in TileDB. Its better to use larger query buffers for TileDB then to half them for double buffering.


Shelnutt2:
@gspowley I had planned to remove the double buffering because we really don't see any benefit from it. However since you've joined I'll hold off on this until you have had time to review the code and get comfortable with TileDB-VCF. We can then both agree to remove this, or if you'd rather keep we can discuss. Due to the thresholds I have set most user queries don't hit double buffer so its not an immediate need to remove it, it just simplifies the code.

gspowley (approved):

PR #346, opened 2021-07-13
TileDB-VCF Hail Integration 

gsvic:


apis/spark/src/main/resources/build-info.properties
```diff
@@ -0,0 +1,8 @@
+[Build Metadata]
```
Shelnutt2, line 1: Should this be included? We probably need to put this in gitignore?

apis/spark/src/main/scala/TileDBVCFHailReader.scala
```diff
@@ -0,0 +1,263 @@
...
+
+
+class TileDBHailVCFReader(var uri: String = null, var samples: Option[String] = Option.empty) extends MatrixHybridReader {
+  override def pathsUsed: Seq[String] = Seq.empty
+  override def columnCount: Option[Int] = None
+  override def partitionCounts: Option[IndexedSeq[Long]] = None
+
+  var df: DataFrame = null
+
+  //  val fmt = vcfReader.fmtAttributes.keySet.toArray.map{ fmtAttr =>
```
Shelnutt2, line 29: Should this be removed?

apis/spark/src/main/scala/TileDBVCFHailReader.scala
```diff
@@ -0,0 +1,263 @@
...
+    if (df == null) {
+      df = ctx.backend.asSpark("spark")
+        .sparkSession.read
+        .format("io.tiledb.vcf")
+        .option("samples", samples.get)
+        .option("uri", uri)
+        .load()
+    }
+
+    //    var projection = df.groupBy("contig","posStart", "alleles")
```
Shelnutt2, line 114: Remove?

apis/spark/src/main/scala/TileDBVCFHailReader.scala
```diff
@@ -0,0 +1,263 @@
...
+        idx += 1
+      }
+    }
+
+    val hasRsid = rowType.hasField("rsid")
+    val hasQual = rowType.hasField("qual")
+    val hasFilter = rowType.hasField("filters")
+
+    // Parse info entry fields
+    val infoFields = Seq("AD", "DP", "GQ", "GT", "MIN_DP", "PL", "SB")
```
Shelnutt2, line 151: There can be a lot more info fields than these, we should use the API we have to list info fields.

apis/spark/src/main/scala/TileDBVCFHailReader.scala
```diff
@@ -0,0 +1,263 @@
...
+        .option("uri", uri)
+        .load()
+    }
+
+    //    var projection = df.groupBy("contig","posStart", "alleles")
+    //      .agg(functions.collect_list("genotype").as("GT"),
+    //        functions.collect_list("fmt_DP").as("DP"),
+    //        functions.collect_list("fmt_PL").as("PL"))
+
+    var projection = df
```
Shelnutt2, line 119: Why are all these fields hard coded?

apis/spark/src/main/scala/TileDBVCFHailReader.scala
```diff
@@ -0,0 +1,263 @@
...
+      .withColumnRenamed("fmt_AD", "AD")
+      .withColumnRenamed("fmt_DP", "DP")
+      .withColumnRenamed("fmt_GQ", "GQ")
+      .withColumnRenamed("genotype", "GT")
+      .withColumnRenamed("fmt_MIN_DP", "MIN_DP")
+      .withColumnRenamed("fmt_PL", "PL")
+      .withColumnRenamed("fmt_SB", "SB")
+
+
+    var columns = Seq[String]("contig", "posStart", "alleles")
```
Shelnutt2, line 129: Why are thee columns hard coded?

Shelnutt2 (changes requested):
Some questions and comments. Also the test dataset must be put into a versioned folder like all other test data so we know which VCF schema the data was created with and as we change overtime we can maintain tests for backwards compatibility.

gspowley:
Closing stale PR.

PR #348, opened 2021-07-23
Add factor to increase the size of fmt/info buffers

Shelnutt2:
Add factor to increase the size of fmt/info buffers


gspowley:
Closing stale PR.

PR #353, opened 2021-08-09
Introduce spdlog for finer grain log verbosity

gspowley:
Add spdlog support for finer grain log verbosity. This release adds `--log-level` and `--log-file` args to the `tiledbvcf` CLI tools.

### Log Level 
Log level is controlled with the `-ll, --log-level <level>` argument:

level | Log messages displayed
:---: | :---
FATAL \| F \| f | LOG_FATAL
ERROR \| E \| e | LOG_ERROR (and above)
WARN \| W \| w | LOG_WARN (and above)
INFO \| I \| i | LOG_INFO (and above)
DEBUG \| D \| d | LOG_DEBUG (and above)
TRACE \| T \| t | LOG_TRACE (and above)

### Log File
If the  `-lf, --log-file <filename>` argument is provided, log messages are also written to `filename`.
* If `filename` exists, log message are appended to `filename`.
* If `filename` includes non-existing directories, the directories are created.

### Code Examples

Add this include and namespace to use logging functions.
```c++
#include "utils/logger_public.h"

using namespace tiledb::vcf;
```

Set the log level and log file with `LOG_CONFIG`.
```c++
LOG_CONFIG(log_level, log_file);
```

Log messages can be strings or formatted with the `fmt` library.
```c++
LOG_INFO(
  "Recorded {} cells for contig {} (task {} / {})",
  worker->records_buffered(),
  contig,
  i + 1,
  tasks.size());
```
```
[2021-08-11 16:37:23.569] [tiledb-vcf] [Process: 821574] [Thread: 821574] [info] Recorded 20 cells for contig 1 (task 1 / 16)
```

`fmt` also supports locale specific formatting.
```c++
LOG_INFO(fmt::format(
  std::locale(""),
  "Done. Ingested {:L} records (+ {:L} anchors) from {:L} samples in {} seconds.",
  records_ingested,
  anchors_ingested,
  samples.size(),
  utils::chrono_duration(start_all)));
```

```
[2021-08-11 16:37:24.007] [tiledb-vcf] [Process: 821574] [Thread: 821574] [info] Done. Ingested 603 records (+ 29,085 anchors) from 3 samples in 0.49614039 seconds.
```


libtiledbvcf/cmake/Modules/FindSpdlog_EP.cmake
```diff
@@ -0,0 +1,70 @@
...
+  REQUIRED_VARS SPDLOG_INCLUDE_DIR
+)
+
+if (NOT SPDLOG_FOUND AND SUPERBUILD)
+  message(STATUS "Adding Spdlog as an external project")
+  ExternalProject_Add(ep_spdlog
+    PREFIX "externals"
+    # Set download name to avoid collisions with only the version number in the filename
+    DOWNLOAD_NAME ep_spdlog.zip
+    URL "https://github.com/gabime/spdlog/archive/v0.16.3.zip"
```
Shelnutt2, line 52: Can we use spdlog 1.9 and switch to cmake targets instead of the header only interface?

Shelnutt2:
The latest build failure is because spdlog dependency for TileDB is finding the spdlog dependency built by TileDB-VCF [1], and TileDB 2.3 is on the old version of spdlog and does not have the fixes for the `error: ‘stdout_color_mt’ is not a member of ‘spdlog’`. The easiest way to address this until TileDB 2.4 is to set the cmake DEPENDS to make sure we build TileDB first then build the spdlog dependency. 

[1] https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=14134&view=logs&j=003489bf-034b-542b-22c5-b80a4fa7d9a2&t=ef4f3ffe-ebfc-5673-51f8-e22bae575e9d&l=8110
[2] https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=14134&view=logs&j=003489bf-034b-542b-22c5-b80a4fa7d9a2&t=ef4f3ffe-ebfc-5673-51f8-e22bae575e9d&l=8185

ihnorton:
👍 . Seems like we are duplicating code from core, so would be good to eventually reuse/centralize the logging helper headers (not necessarily now though).

gspowley:
Agreed, there are some WIP updates to the `Logger` class that may be useful in the core, like optionally logging to a file.

libtiledbvcf/cmake/Modules/FindSpdlog_EP.cmake
gspowley: Yes, we're using spdlog 1.9 now. Thanks for pointing out the TileDB/spdlog version dependency.

libtiledbvcf/src/utils/logger.cc
```diff
@@ -0,0 +1,198 @@
...
+bool Logger::debug_enabled() {
+  return (level_ == spdlog::level::debug) || (level_ == spdlog::level::trace);
+}
+
+/* ********************************* */
+/*              GLOBAL               */
+/* ********************************* */
+
+Logger& global_logger() {
+  static Logger l;
```
Shelnutt2, line 148: Instead of having a single global logger at the process level could we have loggers set on the dataset/reader/writer level or similar? With spark usage it'd be nice to not have a single logger but have separate ones were we eventually have tags such as partition ids etc so its clear which logs go to which query.

Shelnutt2: This doesn't have to be done in this initial version, just a though.

gspowley: Yes, that makes sense. Maybe this PR can cover CLI logging and we can target multiple loggers in the next version.

Shelnutt2: Sounds good

libtiledbvcf/src/c_api/tiledbvcf.cc
```diff
@@ -32,13 +32,16 @@
...
 #include "read/reader.h"
+#include "utils/logger_public.h"
 #include "utils/utils.h"
 #include "write/writer.h"
 
 #include <cassert>
 #include <iostream>
 #include <memory>
 
+using namespace tiledb::common;
```
Shelnutt2, line 43: Why are we using this namespace here? This is an internal namespace to TileDB

gspowley: That namespace came with the `logger.*` files from the core. I'll change the namespace in this repo we don't use the TileDB internal namespace.

Shelnutt2 (approved):

libtiledbvcf/src/utils/logger.cc
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+#include "utils/logger.h"
+
+#include <spdlog/fmt/fmt.h>
+#include <spdlog/fmt/ostr.h>
+#include <spdlog/sinks/basic_file_sink.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+
+namespace tiledb {
```
Shelnutt2, line 41: We should probably have a general discussion on C++ namespaces. Everything in this project is under `tiledb::vcf` currently. I think we should avoid using the `tiledb` namespace which is reserved for the main library. However I don't have super strong opinions. The C++ API for TileDB is under `tiledb` namespace so there is a small potential (although unlikely) for collisions.

gspowley: Good point. I moved the logger code into `tiledb::vcf`. We can revisit the namespace in the future if we decide to share logger code with TileDB core.

PR #377, opened 2021-09-29
Migrate to the CLI11 argument parsing library

gspowley:
Update the command line tool to use [CLI11](https://github.com/CLIUtils/CLI11) for argument parsing.

The `CLI11` library provides robust argument parsing, useful error messages, and nicely organized help messages.


libtiledbvcf/src/write/writer.h
```diff
@@ -55,7 +55,7 @@ struct IngestionParams {
   std::string log_file;
   std::string samples_file_uri;
   std::vector<std::string> sample_uris;
-  unsigned num_threads = std::thread::hardware_concurrency();
+  unsigned num_threads = std::thread::hardware_concurrency() / 2;
```
Shelnutt2, line 58: Should this change be included?

gspowley: Good point, that changes more behavior than just swapping out the CLI library. 

We may decide to change the default number of reader threads, but we can make that change in another PR.

Shelnutt2 (approved):
Nice improvements to the CLI!

PR #388, opened 2021-11-01
Change TSV export prefix for FORMAT fields

gspowley:
Change the TSV export prefix for `FORMAT` fields from `S:`  to `F:`.

[sc-11687]


Shelnutt2 (approved):
Anyway we can support both for backwards compatibility?

Shelnutt2 (approved):

PR #390, opened 2021-11-13
Throw exception when trying to ingest invalid VCF files

gspowley:
* Throw an exception when trying to ingest a VCF file without a valid header.
* Add `try/catch` block in `tiledbvcf` CLI and display critical log message instead of aborting with core dump.

### Old behavior
```shell
$ tiledbvcf store -u test bad.vcf
terminate called after throwing an instance of 'std::runtime_error'
  what():  Invalid VCF file: bad.vcf
Aborted (core dumped)
```

### New behavior
```shell
$ tiledbvcf store -u test bad.vcf
[2021-11-13 13:02:38.506] [tiledb-vcf] [Process: 826246] [Thread: 826246] [critical] Exception: Invalid VCF file: bad.vcf
```


Shelnutt2 (changes requested):
We support sample-less VCF since #366

gspowley:
In that case, we will change this PR to check for invalid VCF files.

Shelnutt2 (approved):

PR #394, opened 2021-11-16
Update to TileDB 2.5.2

aaronwolen:


libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -2197,17 +2198,25 @@ void TileDBVCFDataset::preload_data_array_non_empty_domain() {
...
+  /*
   TRY_CATCH_THROW(
       data_array_preload_non_empty_domain_thread_ = std::async(
           std::launch::async,
           [this]() { data_array_->non_empty_domain<uint32_t>(0); }));
+  */
 }
 
 void TileDBVCFDataset::preload_data_array_non_empty_domain_v4() {
+  data_array_->non_empty_domain_var(0);
```
Shelnutt2, line 2212: Do we need to not async load the non empty domains? I think its only the fragment_info that cause the problem temporarily?

gspowley: Right, I used a big hammer and removed all async preloading. We can scale the change back to only load the fragment_info synchronously.

Shelnutt2 (approved):

PR #399, opened 2021-12-01
Set TileDB total memory budget

gspowley:
Set `sm.mem.total_budget` for read queries. The default is 10GB, but we can get better performance by providing more memory when available.

### DIFF
--- a/libtiledbvcf/src/read/reader.cc
+++ b/libtiledbvcf/src/read/reader.cc
@@ -2420,25 +2420,30 @@ void Reader::set_tiledb_query_config() {
       params_.tiledb_config_map.end())
     cfg["sm.skip_est_size_partitioning"] = "true";
 
+  auto tiledb_total = params_.memory_budget_breakdown.tiledb_tile_cache +
+                      params_.memory_budget_breakdown.tiledb_memory_budget;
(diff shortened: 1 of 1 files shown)


libtiledbvcf/src/read/reader.cc
```diff
@@ -2420,25 +2420,30 @@ void Reader::set_tiledb_query_config() {
       params_.tiledb_config_map.end())
     cfg["sm.skip_est_size_partitioning"] = "true";
 
+  auto tiledb_total = params_.memory_budget_breakdown.tiledb_tile_cache +
+                      params_.memory_budget_breakdown.tiledb_memory_budget;
+  if (!params_.tiledb_config_map.count("sm.mem.total_budget") && tiledb_total) {
```
Shelnutt2, line 2425: @gspowley @KiterLuc is the `sm.mem.total_budget` inclusive of the tile cache? I thought the tile cache was still separate in the new sparse unordered reader?

KiterLuc: @Shelnutt2 we disable the cache now so we can give it all to total-budget

gspowley: Thanks @KiterLuc, I'll make that change.

Shelnutt2 (approved):

PR #426, opened 2022-03-25
Create variant stats array during ingestion

gspowley:
This PR adds a feature to create a TileDB array of variant stats at ingestion time.

* The variant stats array contain data that is used to efficiently compute population wide statistics for each variant (allele count, allele frequency, hom_ref, het, hom_var, etc.). 
* Creating these stats during ingestion saves the cost of reading all the variants after ingestion to create the same stats. 
* Updating the population wide variant stats after ingesting new samples is very efficient because the stats are computed from existing data in the variant stats array. We do not need to read all the existing VCF records again after adding new samples.


gspowley:
One more change to add is submitting the TileDB query inside `VariantStats` to reduce buffer memory usage.

libtiledbvcf/src/dataset/variant_stats.cc
```diff
@@ -0,0 +1,325 @@
...
+//===================================================================
+
+void VariantStats::create(
+    Context& ctx, std::string root_uri, tiledb_filter_type_t checksum) {
+  LOG_DEBUG("VariantStats: Create array");
+
+  // Create filters
+  Filter zstd(ctx, TILEDB_FILTER_ZSTD);
+
+  FilterList filters_1(ctx);
```
Shelnutt2, line 44: Can we name this coords/offsets?

libtiledbvcf/src/dataset/variant_stats.cc
```diff
@@ -0,0 +1,325 @@
...
+  FilterList filters_2(ctx);
+  filters_1.add_filter(zstd);
+  filters_2.add_filter({ctx, TILEDB_FILTER_DOUBLE_DELTA});
```
Shelnutt2, line 47: For offsets I think we should also add bitwidth reduction

libtiledbvcf/src/dataset/variant_stats.cc
```diff
@@ -0,0 +1,325 @@
...
+  // Create schema and domain
+  ArraySchema schema(ctx, TILEDB_SPARSE);
+  schema.set_order({{TILEDB_ROW_MAJOR, TILEDB_ROW_MAJOR}});
+  schema.set_allows_dups(true);
+
+  Domain domain(ctx);
+  const uint32_t pos_min = 0;
+  const uint32_t pos_max = std::numeric_limits<uint32_t>::max() - 1;
+  const uint32_t pos_extent = pos_max - pos_min + 1;
+  auto contig = Dimension::create(
```
Shelnutt2, line 63: Do we want to use the new RLE for contig?

libtiledbvcf/src/dataset/variant_stats.cc
```diff
@@ -0,0 +1,325 @@
...
+    return;
+  }
+
+  std::lock_guard<std::mutex> lock(query_lock_);
+  LOG_DEBUG("VariantStats: Finalize query with {} records", contig_records_);
+  contig_records_ = 0;
+  query_->finalize();
+
+  // Write fragment uri -> sample names to array metadata
+  auto frag_num = query_->fragment_num();
```
Shelnutt2, line 123: I'm not sure if these APIs work with TileDB Cloud, need to followup afterwards.

libtiledbvcf/src/write/writer.cc
```diff
@@ -107,15 +108,27 @@ void Writer::init(const IngestionParams& params) {
 
   // Set htslib global config and context based on user passed TileDB config
   // options
-  utils::set_htslib_tiledb_context(params.tiledb_config);
+  std::vector<std::string> vfs_config = params.tiledb_config;
+  try {
+    auto vcf_region = tiledb_config_->get("vcf.s3.region");
+    vfs_config.push_back("vfs.s3.region=" + vcf_region);
+    LOG_INFO("VFS and htslib reading data from S3 region: {}", vcf_region);
+  } catch (...) {
```
Shelnutt2, line 116: Do we want to log or anything with the exceptions here?

libtiledbvcf/src/dataset/variant_stats.cc
gspowley: Yes, added more descriptive names to the filter lists.

libtiledbvcf/src/dataset/variant_stats.cc
gspowley: Added.

libtiledbvcf/src/dataset/variant_stats.cc
gspowley: Yes, added RLE for the contig dimension. RLE does a great job here.

libtiledbvcf/src/write/writer.cc
gspowley: The try/catch here is used to check if `tiledb_config_` contains `vcf.s3.region`, so no action is required if we catch (I added a comment to make that clear). I didn't find a function to check if a config contains a value.

Shelnutt2: Thanks for clarifying. This might be a gap in the c++ api, feel free to open a story (or PR) for this.

Shelnutt2 (approved):

PR #432, opened 2022-05-09
Disable non-empty domain preload

gspowley:
This PR disables preloading of non-empty domains, which resolves intermittent segfaults seen when opening large VCF datasets on s3.


Shelnutt2 (approved):
We did this to reduce the time it takes for TileDB Cloud Arrays. This likely will cause slowdowns for some of our examples but a segfault is worse. lgtm for now.

PR #437, opened 2022-05-18
Add support for the new TileDB group API

gspowley:
Add support for the new group API in TileDB 2.8.

Example usage with TileDB Cloud and S3:

1. Register a new VCF dataset group with TileDB Cloud:
```
tiledbvcf create -u tiledb://<NAMESPACE>/s3://<BUCKET>/<PREFIX> ...
```

2. Ingest samples to S3 storage:
```
tiledbvcf store -u s3://<BUCKET>/<PREFIX> ...
```

3. Query the VCF dataset group in TileDB Cloud:
```
tiledbvcf export -u tiledb://<NAMESPACE>/<GROUP> ...
```

### DIFF
--- a/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
+++ b/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
@@ -30,6 +30,8 @@
 #include <string>
 #include <vector>
 
+#include <tiledb/tiledb_experimental>  // for the new group api
+
 #include "base64/base64.h"
 #include "dataset/tiledbvcfdataset.h"
 #include "dataset/variant_stats.h"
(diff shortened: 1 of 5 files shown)


Shelnutt2 (changes requested):
I think we also need to change the `open` functionality. I believe also we should add names for the different group members so in the open you can use `root_group.member("data")` etc.

gspowley:
The array `open` functionality is updated to use group member URIs.

gspowley:
The array `open` functionality is updated to use group member URIs.

aaronwolen (approved):
🚀

Shelnutt2 (approved):

PR #445, opened 2022-07-22
fix search for Catch headers

awenocur:
This fix should allow CMake to find an installed Catch 2 header library.


gspowley (approved):
:+1:

Shelnutt2:
Please validate how TileDB core library handles catch2. We should have the same approach in all the repos.

PR #454, opened 2022-08-16
fix: [sc-20370] Add null terminator to string INFO fields in combine VCFs

awenocur:
fix by @gspowley

Story details: https://app.shortcut.com/tiledb-inc/story/20370


libtiledbvcf/src/vcf/vcf_merger.cc
```diff
@@ -352,18 +352,23 @@ void VCFMerger::finish_info(SafeBCFRec& rec) {
...
       if (number == BCF_VL_FIXED || number == BCF_VL_VAR) {
         // merge first value seen in sample order
         if (md_.info[key].size() == 0) {
-          for (int j = 0; j < values; j++) {
-            md_.info[key].push_back(*data++);
+          if (type == BCF_HT_STR) {
+            // bcftools uses strlen to determine the length of BCF_HT_STR fields
+            // For type BCF_HT_STR, values_read is the number of bytes and
+            //   md_.info[key] is a vector of uint32_t (4 bytes)
+            // Allocate enough space to hold values_read bytes + an extra 4
```
Shelnutt2, line 364: You are only adding `+2` here. Also why do we need 4 bytes for the null terminator?

awenocur: Note that for the `+2`, those aren't two bytes, but two 32-bit words, as this is a `std::vector<int>`. There is a minimum of one extra 32-bit word being added at the end, hence the `+2` is to ensure that there still be a full null cell terminator when the number of characters is not a multiple of 4. We may get by just having a `+1`, which I could test, and would still allow for a single byte terminator. In the event that it requires a full null 4-byte element, a ternary expression could be used to limit it to one in the case there be no remainder on that integer division.

awenocur: I also would consider changing the type of `format_keys` to `int32_t`, as we're making an assumption about the width of `int` here that isn't assured according to the language standard.

gspowley: Updated the code to use `utils::ceil`, which is cleaner.

gspowley (approved):

PR #456, opened 2022-10-04
Check for an exact sample range match when resuming ingestion

gspowley:
The resume ingestion logic was changed in #434 to check for an overlap between sample names in the batch being ingested and sample names in existing fragments. This approach does not work when sample names are not sorted.

This PR checks for an exact match between the ingested batch's sample range and the sample range of all existing fragments. If a match is found, and the contig is contained in an existing fragment for the sample range, the contig is skipped.

### DIFF
--- a/libtiledbvcf/src/write/writer.cc
+++ b/libtiledbvcf/src/write/writer.cc
@@ -334,7 +334,7 @@ void Writer::ingest_samples() {
   if (ingestion_params_.resume_sample_partial_ingestion) {
     LOG_DEBUG(
         "Starting fetching of contig to sample list for resumption checking");
-    existing_fragments = dataset_->fragment_contig_sample_list();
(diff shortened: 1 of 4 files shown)


Shelnutt2 (approved):

libtiledbvcf/src/write/writer.cc
```diff
@@ -780,48 +780,51 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples_v4(
...
         }
-        if (skip) {
-          break;
-        }
       }
 
       // Remove the region if marked to skip
       if (skip) {
-        LOG_DEBUG("Resume: skipping contig {}", contig);
+        LOG_DEBUG("Resume:   skipping contig {}", contig);
```
awenocur, line 821: Is this meant to be indented?

awenocur (approved):

PR #469, opened 2022-11-30
Add zstd compression to the sample dimension

gspowley:
Provide a `tiledbvcf create` option to use dictionary + zstd compression on the sample dimension, instead of dictionary encoding only. This option improves the compression ratio.

```
  --compress-sample-dim                 Add zstd compression to the sample dimension.
```


awenocur (approved):

Shelnutt2 (commented):
I believe we should make this optional or configurable by the user. There are upcoming core changes to continue to take advantage of the dictionary encoded data l, such as working on the encoded data directly.

gspowley:
We will make filters configurable by the user in sc-23728. This PR provides the option to use dictionary + zstd on the sample dim (instead of dictionary only) until sc-23728 is finished.

Shelnutt2 (approved):

PR #489, opened 2023-02-02
Memory usage and latency improvements

gspowley:
* Reduce memory usage when fetching the one VCF header.
* Enable async mode for AF filtering to allow the variant stats query to run in parallel with the VCF data query.
* Add developer convenience `Makefile` for common commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,105 @@
+# This Makefile captures common developer build and test use cases.
+
+MAKEFLAGS += --no-print-directory
+
+# print help by default
+help:
+
+# install 
+# -------------------------------------------------------------------
+
+# set default variable values, if non-null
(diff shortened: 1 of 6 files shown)


awenocur (approved):

Makefile
```diff
@@ -0,0 +1,105 @@
+# This Makefile captures common developer build and test use cases.
```
Shelnutt2, line 1: I don't think we should mix make and cmake. @ihnorton do you have thoughts here? We don't mix things in any other repo.

gspowley: This idea came from SOMA, which does have a Makefile. It's a convenience Makefile to run common developer commands, so people don't have to come up with their own aliases, scripts, etc.

PR #495, opened 2023-02-03
Release python GIL

gspowley:
Release the python GIL, to allow other python threads to run while libtiledbvcf is reading/writing TileDB arrays.

### DIFF
--- a/apis/python/src/tiledbvcf/binding/reader.cc
+++ b/apis/python/src/tiledbvcf/binding/reader.cc
@@ -225,6 +225,9 @@ void Reader::read(const bool release_buffs) {
   alloc_buffers(release_buffs);
   set_buffers();
 
+  // Release python GIL after we're done accessing python objects.
+  py::gil_scoped_release release;
+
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbvcf/binding/reader.cc
```diff
@@ -225,6 +225,9 @@ void Reader::read(const bool release_buffs) {
   alloc_buffers(release_buffs);
   set_buffers();
 
+  // Release python GIL after we're done accessing python objects.
+  py::gil_scoped_release release;
```
ihnorton, line 229: I would suggest putting this in its own inner scope block around only the expensive functions, to reduce the risk of eg accidentally adding some pybind11 object with function lifetime above in the future.

ihnorton (approved):
LGTM except for minor nit.

apis/python/src/tiledbvcf/binding/reader.cc
gspowley: Good idea, added the inner scope block.

awenocur (approved):

PR #504, opened 2023-02-27
Handle one missing allele in stats arrays

gspowley:
During stats ingestion, handle the case where one allele is missing, for example GT=".|1" or GT="1|.".


ckrilow (approved):
These changes look good to me. The quick turn around time is much appreciated.

PR #513, opened 2023-03-16
Smr/sc 24610/test remote global order write cache

awenocur:


libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -317,17 +327,38 @@ void AlleleCount::flush() {
...
+        LOG_FATAL("AlleleCount: error submitting TileDB write query");
+      }
     }
 
     // Insert sample names from this query into the set of fragment sample names
     fragment_sample_names_.insert(sample_names_.begin(), sample_names_.end());
     sample_names_.clear();
   }
 
+  // For remote global order writes, zero query buffers prior to
```
Shelnutt2, line 347: Why are we zeroing buffers? The last submit call should be `submit_and_finalize`?

libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -317,17 +327,38 @@ void AlleleCount::flush() {
...
+    query_
+        ->set_buffer(
+            COLUMN_NAME[CONTIG], contig_offsets_.data(), 0, &contig_buffer_, 0)
+        .set_buffer(COLUMN_NAME[POS], pos_buffer_.data(), 0)
+        .set_buffer(COLUMN_NAME[REF], ref_offsets_.data(), 0, &ref_buffer_, 0)
+        .set_buffer(COLUMN_NAME[ALT], alt_offsets_.data(), 0, &alt_buffer_, 0)
+        .set_buffer(
+            COLUMN_NAME[FILTER], filter_offsets_.data(), 0, &filter_buffer_, 0)
+        .set_buffer(COLUMN_NAME[GT], gt_offsets_.data(), 0, &gt_buffer_, 0)
+        .set_buffer(COLUMN_NAME[COUNT], count_buffer_.data(), 0);
```
gspowley, line 359: These buffers are cleared by the `clear` statements below. We can reuse the `set_data_buffer` /`set_offsets_buffer` code above after the buffers are cleared, and it doesn't need to depend on `remote_`.

libtiledbvcf/src/stats/allele_count.h
```diff
@@ -230,6 +230,9 @@ class AlleleCount {
   // Sample names included in the fragment
   inline static std::set<std::string> fragment_sample_names_;
 
+  // Remote flag, true when the VCF dataset is remote.
+  inline static bool remote_ = false;
```
gspowley, line 234: Based on previous comments, can we remove the need to track if the array is remote?

libtiledbvcf/src/stats/variant_stats.cc
gspowley, line 1: Same comments as `allele_count.cc`.

libtiledbvcf/src/stats/variant_stats.h
gspowley, line 1: Same comments as `allele_count.h`.

libtiledbvcf/src/write/writer.cc
```diff
@@ -454,7 +456,11 @@ void Writer::ingest_samples() {
     result = ingest_samples(ingestion_params_, local_samples, regions);
 
     // Make sure to finalize for v2/v3
-    query_->finalize();
+    if (remote_) {
+      query_->submit_and_finalize();
+    } else {
+      query_->finalize();
```
gspowley, line 462: We can call `submit_and_finalize` in all cases if the query buffers are empty.

libtiledbvcf/src/write/writer.cc
```diff
@@ -608,11 +614,27 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples(
...
+            if (query_->query_status() == Query::Status::FAILED) {
+              LOG_FATAL(
+                  "Error submitting TileDB write query; unexpected query "
+                  "status: {}",
+                  // TODO: fix for SC-26607
+                  static_cast<int>(query_->query_status()));
+            }
+            worker->buffers().clear_query_buffers(
+                query_.get(), dataset_->metadata().version);
+          } else {
```
gspowley, line 629: Can we rewrite this status check so it doesn't depend on tracking is the array is remote?

libtiledbvcf/src/write/writer.cc
```diff
@@ -971,6 +993,10 @@ std::pair<uint64_t, uint64_t> Writer::ingest_samples_v4(
                 // Write and finalize anchors stored in the anchor worker.
                 anchors_ingested += write_anchors(anchor_worker);
 
+                if (remote_) {
```
gspowley, line 996: Always clear buffers?

libtiledbvcf/src/write/writer.cc
```diff
@@ -1356,7 +1410,11 @@ void Writer::dataset_version(int32_t* version) const {
 }
 
 void Writer::finalize_query(std::unique_ptr<tiledb::Query> query) {
-  query->finalize();
+  if (remote_) {
+    query->submit_and_finalize();
+  } else {
+    query->finalize();
```
gspowley, line 1416: We can call `submit_and_finalize` if the buffers are cleared. We should add a check to ensure the buffers are empty.

gspowley (commented):
This looks good. Comments propose refactoring to remove the need to track if the array is remote.

libtiledbvcf/src/write/writer.cc
shaunrd0: All done here :tada:  I reran validation today for ingesting the full VCF dataset using instructions on [SC-24610](https://app.shortcut.com/tiledb-inc/story/24610/test-tile-aligned-write-caching-pr-3762-with-tiledb-vcf) and checksums are still good. If we can validate allele count or variant stats I can check those as well.

I had to be a bit more selective on when we zero buffers than I initially thought. If we set the query buffers with size of 0 after every submit and ingestion finds there's still more data to write, then we end up hitting [this](https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/query/query.cc#L1352) error in core if we attempt to re-set the previously zeroed buffers. So we only zero query buffers just before we finalize, and not after every submit.

gspowley: Thanks @shaunrd0, this is great work!

gspowley (approved):

PR #526, opened 2023-05-02
Change the metadata group to a directory

gspowley:
The metadata directory is needed to preserve the local, on-disk V4 directory structure, but it no longer needs to be a group.

### DIFF
--- a/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
+++ b/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
@@ -305,11 +305,11 @@ void TileDBVCFDataset::create_empty_metadata(
     const std::string& root_uri,
     const Metadata& metadata,
     const tiledb_filter_type_t& checksum) {
-  // If the dataset is local, create the metadata group to preserve the V4
-  // directory structure. We do not create the group on TileDB Cloud because it
-  // is unused and will create an orphaned group.
+  // If the dataset is local, create the metadata directory to preserve the V4
+  // directory structure.
   if (!cloud_dataset(root_uri)) {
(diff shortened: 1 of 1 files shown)


gspowley:
This was a low effort/risk change to fix the issue. We could use VFS to create the required directory when ingesting to a local disk. I'll look into that.

Shelnutt2:
> This was a low effort/risk change to fix the issue. We could use VFS to create the required directory when ingesting to a local disk. I'll look into that.

I believe it's important we keep a common and universal on disk data structure, else we will end up with mismatched and incompatible datasets.

Also if f we remove the metadata group, should we bump the version since this group will no longer exist?

gspowley:
We do have a common on disk data structure and it continues to be consistent with this PR. 

The metadata "group" has been empty since we added support for the TileDB group API in #437. The empty metadata group is only there to preserve `metadata` directory in the V4 on disk format.

This PR is addressing an issue we see when ingesting VCF with `tiledb://` URIs. In this case, when the empty metadata group is created, it is registered on TileDB Cloud, but it is not associated with the VCF dataset, it is "orphaned" (maybe a TileDB Cloud issue).

Anyway, we don't want the empty metadata group to be registered on TileDB Cloud. Some options:
1. Don't create the group on TileDB Cloud (the first version of this PR)
2. Create the `metadata` directory with VFS instead of an empty group (the current version of this PR)

Shelnutt2:
If you take option 1, we will end up in an inconsistent state right? If I create it on s3 there is an extra metadata group. If I register the s3 created dataset we'll get a metadata group. This means we'll have basically two on disk structures, which we want to avoid.

PR #530, opened 2023-05-10
Minor updates to python API

gspowley:
* Update some default values in the python API to match the defaults in C++, so we get the desired defaults.
* Add convenience arg `tiledb_config` to `Dataset` to allow passing a TileDB config without creating a `ReadConfig` object.
* Minor comment updates and code refactoring.
* Fix the doc landing page.

### DIFF
--- a/apis/python/src/tiledbvcf/dataset.py
+++ b/apis/python/src/tiledbvcf/dataset.py
@@ -94,6 +94,8 @@ class Dataset(object):
         Enable internal TileDB statistics.
     verbose
         Enable verbose output.
+    tiledb_config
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbvcf/dataset.py
```diff
@@ -103,7 +105,14 @@ def __init__(
         cfg: ReadConfig = None,
         stats: bool = False,
         verbose: bool = False,
+        tiledb_config: dict = None,
     ):
+        if cfg and tiledb_config:
```
Shelnutt2, line 110: Very nice!

PR #537, opened 2023-06-16
Update to c++17 in setup.py

gspowley:
Avoid compile errors seen with some versions of python.


awenocur (approved):

Shelnutt2:
@gspowley can we rebase this and rerun to check the CI failure? Supporting c++17 would let us also use newer arrow versions.

PR #538, opened 2023-06-21
Update to C++ accelerated IAF computation

awenocur:


apis/python/tests/test_tiledbvcf.py
```diff
@@ -1168,6 +1168,18 @@ def test_ingest_with_stats(tmp_path):
         data_frame[data_frame["sample_name"] == "second"]["info_TILEDB_IAF"].iloc[0][0]
         == 0.0625
     )
+    data_frame = ds.read(
+        samples=sample_names,
+        attrs=["contig", "pos_start", "id", "qual", "info_TILEDB_IAF", "sample_name"],
+        scan_all_samples=True,
+    )
+    print(f"bad data frame: {data_frame}")
```
gspowley, line 1176: Was this for debugging?

libtiledbvcf/src/read/reader.cc
```diff
@@ -1282,8 +1283,16 @@ bool Reader::process_query_results_v4() {
 
       read_state_.query_results.af_values.clear();
       int allele_index = 0;
+      std::vector<std::vector<char>>::size_type num_samples = 0;
+      if (params_.scan_all_samples)
```
gspowley, line 1287: Add brackets.

libtiledbvcf/src/read/reader.cc
```diff
@@ -1282,8 +1283,16 @@ bool Reader::process_query_results_v4() {
 
       read_state_.query_results.af_values.clear();
       int allele_index = 0;
+      std::vector<std::vector<char>>::size_type num_samples = 0;
```
gspowley, line 1286: How about an `int` or `uint` type to make the code more readable?

libtiledbvcf/src/stats/variant_stats_reader.h
```diff
@@ -102,6 +103,46 @@ class AFMap {
...
+   *
+   * @param pos Position of the allele
+   * @param allele Allele value
+   * @param num_samples Number of samples in the entire dataset
+   * @return float Allele Frequency
+   */
+  float af(
+      uint32_t pos,
+      const std::string& allele,
+      std::vector<std::vector<char>>::size_type num_samples) {
```
gspowley, line 118: Here too, `int` or `uint`.

libtiledbvcf/src/stats/variant_stats_reader.h
```diff
@@ -102,6 +103,46 @@ class AFMap {
...
+        uint32_t,
+        std::pair<int, std::unordered_map<std::string, int>>>::const_iterator
+        pos_map_iterator = ac_map_.find(pos);
+
+    // Return -1.0 if the allele was not called
+    if (pos_map_iterator == ac_map_.end()) {
+      return -1.0;
+    }
+
+    const std::pair<int, std::unordered_map<std::string, int>>& pos_map =
```
gspowley, line 130: This is begging for an `auto`.

libtiledbvcf/src/stats/variant_stats_reader.h
```diff
@@ -102,6 +103,46 @@ class AFMap {
...
+        pos_map_iterator->second;
+
+    // We don't know that allele was called in this sample. Ask nicely for the
+    // allele count.
+    decltype(pos_map.second)::const_iterator next_allele =
```
gspowley, line 135: Also `auto`.

libtiledbvcf/src/stats/variant_stats_reader.h
```diff
@@ -102,6 +103,46 @@ class AFMap {
...
+        pos_map.second.find(allele);
+
+    // First multiply by 1.0 to force a float type, then look up AC from the
+    // above iterator. Substitute 0 for the AC value if it is absent in pos_map.
+    // Divide by AN (pos_map.first) to get AF.
+    return 1.0 *
+           (next_allele == pos_map.second.end() ? 0 : next_allele->second) /
```
gspowley, line 142: This is where `if/else` is more readable than a ternary operator.

gspowley (changes requested):
Looks good! Let's add code to check the version of the `variant_stats` array, since we need the newer version to compute IAF.

apis/python/tests/test_tiledbvcf.py
awenocur: whoops

libtiledbvcf/src/stats/variant_stats_reader.h
awenocur: this is actually older code, but I can fix it here

libtiledbvcf/src/stats/variant_stats_reader.h
awenocur: This is also where the code is not DRY; would it make sense for me to dedup the `af()` methods?

apis/python/tests/test_tiledbvcf.py
awenocur: the print statement is removed, and the commit message updated; is there anything else to do here?

libtiledbvcf/src/stats/variant_stats_reader.cc
```diff
@@ -37,6 +37,28 @@ VariantStatsReader::VariantStatsReader(
...
+      &variant_stats_version_read);
+  if (version_datatype == TILEDB_UINT32 && version_count == 1) {
+    std::memcpy(
+        &variant_stats_version,
+        variant_stats_version_read,
+        tiledb_datatype_size(version_datatype) * version_count);
+  } else {
+    throw std::runtime_error("Invalid metadata in stats table");
+  }
+  if (variant_stats_version != 2) {
```
gspowley, line 57: Let's throw the error if the version is less than 2 to cover future versions > 2.

libtiledbvcf/src/stats/variant_stats_reader.cc
```diff
@@ -37,6 +37,28 @@ VariantStatsReader::VariantStatsReader(
...
+    throw std::runtime_error(
+        "Variant stats table invoked from dataset with incompatible stats "
```
gspowley, line 59: Let's give an error message that tells the user how to resolve the issue, something like "The variant_stats array version cannot be used to compute IAF. Please re-ingest to enable IAF support".

libtiledbvcf/src/stats/variant_stats_reader.h
gspowley: Yes, it's worth looking at deduping the `af()` methods.

gspowley (approved):
Looks good!

PR #574, opened 2023-09-13
Handle htslib errors reading VCF data

gspowley:
Throw an error if htslib returns an error reading VCF data.

### DIFF
--- a/libtiledbvcf/src/vcf/vcf_v4.cc
+++ b/libtiledbvcf/src/vcf/vcf_v4.cc
@@ -480,10 +480,18 @@ void VCFV4::Iter::swap(VCFV4::Iter& other) {
 bool VCFV4::Iter::next(bcf1_t* rec) {
   int ret;
 
+  // bcf_itr_next and tbx_itr_next return -2 on error
+
   if (tbx_ == nullptr) {
     ret = bcf_itr_next(fh_.get(), hts_iter_, rec);
+    if (ret < -1) {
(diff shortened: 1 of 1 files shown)


libtiledbvcf/src/vcf/vcf_v4.cc
```diff
@@ -480,10 +480,18 @@ void VCFV4::Iter::swap(VCFV4::Iter& other) {
 bool VCFV4::Iter::next(bcf1_t* rec) {
   int ret;
 
+  // bcf_itr_next and tbx_itr_next return -2 on error
+
   if (tbx_ == nullptr) {
     ret = bcf_itr_next(fh_.get(), hts_iter_, rec);
+    if (ret < -1) {
+      throw std::runtime_error("Error reading VCF record");
```
Shelnutt2, line 488: Anyway to get error details?

gspowley: No, htslib only indicates there was an error in the return code.

awenocur: just reviewed the relevant htslib code; there are no errors set apart from return values; can confirm that we're limited to an integer for the error code, and the only information contained therin is a -1 vs. -2 return value.

It's cryptic, but I think -1 means unexpected EOF while -2 means anything else.

awenocur (approved):

PR #577, opened 2023-09-15
Add nightly build (linux only for now)

jdblischak:
Nightly build against nightly libtiledb. Will add macOS and Windows builds in the future. The main blocker is finding a compatible pyarrow. I've tried multiple versions of pyarrow, installed via either pip or conda, and I can't find an option that successfully builds for both Linux and macOS.

cc: @ihnorton, @awenocur

### DIFF
--- a/ci/nightly/build-tiledbvcf-py.sh
+++ b/ci/nightly/build-tiledbvcf-py.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 set -ex
 
-# Build tiledbvcf-py assuming source code directory is ./TileDB-VCF/apis/python
+# Build (and test) tiledbvcf-py assuming source code directory is
+# ./TileDB-VCF/apis/python
 
 export LD_LIBRARY_PATH=$GITHUB_WORKSPACE/install/lib:${LD_LIBRARY_PATH-}
 echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"
@@ -9,3 +10,5 @@ echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"
 cd TileDB-VCF/apis/python
 python setup.py install --libtiledbvcf=$GITHUB_WORKSPACE/install/
 python -c "import tiledbvcf; print(tiledbvcf.version)"
+
+pytest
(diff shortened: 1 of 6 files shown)


awenocur (approved):

ci/nightly/build-tiledbvcf-py.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ex
+
+# Build tiledbvcf-py assuming source code directory is ./TileDB-VCF/apis/python
+
+export LD_LIBRARY_PATH=$GITHUB_WORKSPACE/install/lib:${LD_LIBRARY_PATH-}
```
ihnorton, line 6: Do you know why this is necessary? cc @awenocur. I see there are other usages in the CI setup. This may mean that we are not configuring RPATH correctly.

ihnorton (changes requested):
Thanks. Let's also run the tests for libtiledbvcf and tiledb-py (unless I missed where they are being run). Approved to merge after the testing is added, so that we start getting this coverage ASAP; we'll address the LD_LIBRARY_PATH issue in follow-up.

jdblischak:
> One question: should I also build dev TileDB-Py from source? Since I install TileDB-Py from conda, `ldd` shows it linked to the conda-installed libtiledb (ie not the nightly one I built from source and installed in `./install/`)

I'm less concerned about this now after doing more local testing. While both `tiledb.libtiledb.version()` and `ldd` say that the conda-installed TileDB-Py is linked to the conda-installed 2.17.0, which has S3 support built-in during the build, I still have reason to believe that TileDB-Py is accessing the dev `.so` (2.18.0), presumably because I manually set `LD_LIBRARY_PATH`. When I don't build dev libtiledb with S3 support, the pytest tests fail with `[TileDB::VFS] Error: TileDB was built without S3 support`. And when I re-built libtiledb from source with `-D TILEDB_S3=ON`, then the pytest tests passed.

ihnorton:
Thanks 💯 

> When I don't build dev libtiledb with S3 support, the pytest tests fail with

Nice differential!!

> One question: should I also build dev TileDB-Py from source?
> I'm less concerned about this now after doing more local testing.

We could end up with symbol issues at some point, but as long as it works, lets get this going and address this and other issues in follow-up.

ihnorton (approved):

PR #580, opened 2023-09-21
Enable LIBTILEDBVCF_PATH to be set as an environment variable

jdblischak:
It is easy to pass `--libtiledbvcf` to `python setup.py install`, but this is trickier when trying to build with `pip install`. This PR adds the additional option of specifying the env var `LIBTILEDBVCF_PATH`, though `--libtiledbvcf` takes precedence

This is similar to what is currently done in [TileDB-Py](https://github.com/TileDB-Inc/TileDB-Py/blob/dev/setup.py#L577-L579) and [TileDB-SOMA](https://github.com/single-cell-data/TileDB-SOMA/blob/7a0d937718077310fde128840a972230007b7694/apis/python/setup.py#L43-L60)

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -41,7 +41,7 @@ import sys
 
 TILEDBVCF_DEBUG_BUILD = False
 TILEDBVCF_S3 = True
-LIBTILEDBVCF_PATH = None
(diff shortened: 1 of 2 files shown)


ihnorton (commented):
Approach LGTM

awenocur (approved):

PR #582, opened 2023-09-23
Iterate over polyploid calls

awenocur:
Genotypes are being handled with hard-coded logic for diploid calls. Ensure that this be abstracted out as an iteration.

### DIFF
--- a/libtiledbvcf/src/stats/allele_count.cc
+++ b/libtiledbvcf/src/stats/allele_count.cc
@@ -426,16 +426,6 @@ void AlleleCount::process(
     return;
   }
 
-  // Only haploid and diploid are supported
-  if (ngt > 2) {
-    LOG_FATAL(
-        "Ploidy > 2 not supported: sample={} locus={}:{} ploidy={}",
-        sample_name,
-        contig,
-        pos,
-        ngt);
-  }
-
   // Skip if homozygous ref or alleles are missing
   if (ngt == 1) {
     if (gt0 == 0 || gt0_missing) {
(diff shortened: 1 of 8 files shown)


libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -465,39 +457,40 @@ void AlleleCount::process(
   // Sort ALT alleles and normalize GT
   //  - haploid GT = 1
   //  - diploid GT = 0,1 or 1,1 or 1,2
-  std::string alt;
-  std::string gt;
-  if (ngt == 1) {
-    // haploid
-    alt = rec->d.allele[gt0];
-    gt = "1";
+  std::string sulected_alt;
```
gspowley, line 460: Typo in variable name?

libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -465,39 +457,40 @@ void AlleleCount::process(
...
-    if (gt0_missing || gt1_missing) {
-      gt = ".,1";
-      alt = gt0_missing ? alt1 : alt0;
-    } else if (!gt0 || !gt1) {
-      gt = "0,1";
-      alt = gt0 ? alt0 : alt1;
-    } else if (gt0 == gt1) {
-      gt = "1,1";
-      alt = alt0;
+    selected_gt = "1,2";
```
gspowley, line 478: We haven't defined how ployploid variants should be represented in the allele count array. We could release the variant stats and IAF changes in this PR and follow up with the allele count changes in another PR.

awenocur: There's no logic yet to handle polyploid variants, just a change from a scalar to vector representation in anticipation of the next step.

gspowley (approved):

PR #593, opened 2023-10-05
Sort reads during sample delete

gspowley:
When deleting a sample, the records need to be read in sorted order so the stats arrays can be updated without violating the global order.


awenocur (approved):

gspowley:
[sc-35314]

PR #602, opened 2023-10-24
Allow calls with missing genotypes to pass the IAF filter

gspowley:
This fixes an issue where a variant with all GT values missing (for example, a gVCF block) would not pass the IAF filter.


leipzig:
I'm ok with approving this if a unit test can be earmarked

leipzig (approved):

awenocur (approved):

PR #648, opened 2024-01-09
[Spark] Add option to use a bed Array

DimitrisStaratzis:
This PR follows #646 [sc-38682] and enables the setting of a TileDB bed array in both Spark2 and Spark3. To indicate the correctness of the new feature, new tests have been added


ihnorton (commented):
LGTM but let's fix the CI failures and get a +1 from @gspowley before merging.

gspowley (commented):
LGTM. Ready to merge when CI is passing. Thanks @DimitrisStaratzis!

DimitrisStaratzis:
After a closer look some work needs to be done to enable the bedarray with the `new partition method`. More specifically, I see that the new partition method is using this method:

```java
List<List<String>> computeRegionPartitionsFromBedFile(int desiredNumRangePartitions) {
  Optional<URI> bedURI = options.getBedURI();

  log.info("Init VCFReader for partition calculation");
  String uriString = options.getDatasetURI().get().toString();

  Optional<String> credentialsCsv =
      options
          .getCredentialsProvider()
          .map(CredentialProviderUtils::buildConfigMap)
          .flatMap(VCFDataSourceOptions::getConfigCSV);

  Optional<String> configCsv =
      VCFDataSourceOptions.combineCsvOptions(options.getConfigCSV(), credentialsCsv);

  String[] samples = new String[] {};
  VCFReader vcfReader = new VCFReader(uriString, samples, options.getSampleURI(), configCsv);

  VCFBedFile bedFile = new VCFBedFile(vcfReader, bedURI.get().toString());

  Map<String, List<String>> mapOfRegions = bedFile.getContigRegionStrings();
  List<List<String>> res = new LinkedList<>(mapOfRegions.values());

  // Sort the region list by size of regions in contig, largest first
  res.sort(Comparator.comparingInt(List<String>::size).reversed());

  // Keep splitting the larges region lists until we have the desired minimum number of range
  // Partitions, we stop if the large region has a size of 10 or less
  while (res.size() < desiredNumRangePartitions && res.get(0).size() >= 10) {

    List<String> top = res.remove(0);

    List<String> first = new LinkedList<>(top.subList(0, top.size() / 2));
    List<String> second = new LinkedList<>(top.subList(top.size() / 2, top.size()));
    res.add(first);
    res.add(second);

    // Sort the region list by size of regions in contig
    res.sort(Comparator.comparingInt(List::size));
    Collections.reverse(res);
  }

  bedFile.close();
  vcfReader.close();

  return res;
}

```
So I need to implement

'''java
List<List<String>> computeRegionPartitionsFromBedArray(int desiredNumRangePartitions) {}
```

basically what I need to do here is replace bedFile.getContigRegionStrings(); with something like bedArray.getContigRegionStrings();

File: apis/spark/src/main/java/io/tiledb/vcf/VCFInputPartitionReader.java [hunk 1]
```diff
@@ -304,6 +304,12 @@ private void initVCFReader() {
       if (bedURI.isPresent()) {
         vcfReader.setBedFile(bedURI.get().toString());
       }
+
+      // Set BED array
+      Optional<URI> bedArrayURI = options.getBedArrayURI();
+      if (bedArrayURI.isPresent()) {
+        vcfReader.setBedArray(bedArrayURI.get().toString());
```
gspowley, line 311: We recently removed `vcfReader.setBedArray` to simplify the API. Please use `vcfReader.setBedFile` instead. I believe that's why CI is failing.

gspowley (approved):
Looks good, thanks @DimitrisStaratzis!

DimitrisStaratzis:
CI is green!

PR #672, opened 2024-03-13
More robust handling of info END fields

gspowley:
This PR introduces two changes related to the info `END` field:
1. The info `END` field is normally defined as an integer in VCF headers. This PR handles the case where `END` is defined as a string in a VCF header.
2. The `info_END` attribute in the TileDB array now contains the original value of `END` from the VCF, including missing values.

The `pos_end` TileDB attribute has not changed. If info `END` is defined, `pos_end` holds the value of info `END`. Otherwise, `pos_end` holds the value `POS` + `length(REF)` - 1.


awenocur (approved):

jdblischak:
You can rebase onto main to fix the failing CI. I fixed the dask-expr error in #673

jdblischak:
Something went wrong with the Azure jobs. They aren't even in the [queue](https://dev.azure.com/TileDB-Inc/CI/_build?definitionId=8&_a=summary)

awenocur (approved):

PR #676, opened 2024-03-14
Update to Embedded 2.21.0

awenocur:


ihnorton (changes requested):
I think this is missing the actual version update?

awenocur:
> I think this is missing the actual version update?

whoops; the bump was lost in the latest rebase. It should be back.

ihnorton (approved):

ihnorton:
(FYI for next time: there is not a CSV file attached to the release, with all of the files and hashes)

PR #693, opened 2024-04-04
Do not disable vcpkg in nightly builds.

teo-tsirpanis:


teo-tsirpanis:
Nightlies on dev succeeded.

awenocur (approved):

PR #698, opened 2024-04-10
Writer changes for Variant Stats

awenocur:
[sc-44861]

### DIFF
--- a/libtiledbvcf/src/stats/variant_stats.cc
+++ b/libtiledbvcf/src/stats/variant_stats.cc
@@ -495,11 +495,12 @@ void VariantStats::process(
   bool is_nr_block;
   {
     bool one_alt = rec->n_allele == 2;
+    bool is_ref = bcf_gt_allele(dst_[0]) == 0;
     // if no first alt, or if wrong number of alts, this will be blank:
     // no need to check whether this be a ref block, because this bool will be
     // used inside an if statement
     auto alt = one_alt ? std::string(rec->d.allele[1]) : "";
-    is_nr_block = (alt == "<NON_REF>");
+    is_nr_block = is_ref && (alt == "<NON_REF>");
   }
 
   int length = end_pos - pos + 1;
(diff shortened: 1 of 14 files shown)


awenocur:
Check that the variant stats array version is bumped.

awenocur:
[sc-44861]

libtiledbvcf/src/stats/variant_stats.cc
```diff
@@ -435,7 +456,7 @@ void VariantStats::process(
       return;
     }
   } else {
-    return;  // ngt <= 0
+    return;  // ngt >= 0
```
gspowley, line 459: Let's remove the comment. The return actually happens when `ngt < 0`, which is easy enough to see from the code.

gspowley (commented):
Looks good so far! A couple of more items to complete are captured in shortcut.

libtiledbvcf/src/stats/variant_stats.cc
awenocur: whoops; I misinterpreted that as one of the clang-format generated end bracket comments

libtiledbvcf/src/stats/variant_stats.cc
```diff
@@ -463,16 +492,39 @@ void VariantStats::process(
...
     if (!gt_missing[i]) {
       auto alt = alt_string(ref, rec->d.allele[gt[i]]);
 
       if (gt[i] == 0) {
-        values_["ref"][AC] += count_delta_;
+        std::string ref_key = "ref";
+        if (is_nr_block) {
+          // ref block
+          ref_key = "nr";
+          values_["nr"][CUM_MAX] = max_length_;
```
gspowley, line 519: Is this case, max_length will be the max length of ref blocks. Should we keep track of the longest record, including variants and ref blocks?

libtiledbvcf/src/stats/variant_stats.h
```diff
@@ -206,16 +206,17 @@ class VariantStats {
...
   inline static const std::vector<std::string> COLUMN_STR = {
-      "contig", "pos", "allele"};
+      "contig", "pos", "sample", "allele"};
 
   // Array attributes
-  enum Attr { AC = 0, N_HOM, LAST_ };
-  inline static const std::vector<std::string> ATTR_STR = {"ac", "n_hom"};
+  enum Attr { AC = 0, AN, N_HOM, CUM_MAX, END_POS, LAST_ };
+  inline static const std::vector<std::string> ATTR_STR = {
+      "ac", "an", "n_hom", "cum_max", "end_pos"};
```
gspowley, line 219: Let's rename `cum_max` to `max_length` and `end_pos` to `end`.

gspowley (approved):
Looks good!

PR #700, opened 2024-04-11
Bump nightly release branch to 2.21

jdblischak:
The recent fix for CMake 3.29.1 was backported to [`release-2.21`](https://github.com/TileDB-Inc/TileDB/tree/release-2.21) (https://github.com/TileDB-Inc/TileDB/pull/4864)

Closes #694

PR #669, opened 2024-03-11
Import pyarrow-hotfix

jdblischak:
Until we can install `pyarrow >=14.0.2` in the cloud conda environments, we should ensure that pyarrow-hotfix is installed and imported. From its [PyPI](https://pypi.org/project/pyarrow-hotfix/) Usage section:

> `pyarrow_hotfix` must be imported in your application or library code for it to take effect

xref: https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/115

Also, I noticed that pyarrow isn't included in `setup_requires`

https://github.com/TileDB-Inc/TileDB-VCF/blob/b3df71ea2796c878c3c8c34e3a393accc5cc6ea2/apis/python/setup.py#L284-L292

Which seems unlikely to be true given that pyarrow is imported in `setup.py`:

https://github.com/TileDB-Inc/TileDB-VCF/blob/b3df71ea2796c878c3c8c34e3a393accc5cc6ea2/apis/python/setup.py#L243

In general, is there a reason that `setup_requires`, `install_requires`, and `test_requires` aren't overly utilized in this repo? I expected to see at least pyarrow in `install_requires` and dask in `test_requires`.

PR #701, opened 2024-04-12
Update python build system

gspowley:
The goal of this PR is to update the python build system to the more modern `scikit-build-core`, with inspiration from [TileDB-Inc/TileDB-Vector-Search](https://github.com/TileDB-Inc/TileDB-Vector-Search).

Looking for feedback from build experts, especially related to the current Windows failure and releasing to pypi in the future.


jdblischak:
> related to the current Windows failure

I think the Windows failure is due to the loss of the flag `--libtiledbvcf`. The Windows build setup is similar to that used by [tiledb-vcf-feedstock](https://github.com/TileDB-Inc/tiledb-vcf-feedstock/tree/761620c8be91d46be834f2ff4661a40325212ad7/recipe). The C++ library is installed into the conda environment, and then the Python client finds it via `--libtiledbvcf` (note that in the feedstock the linux and osx builds also use this strategy).

TileDB-Vector-Search and [tiledb-vector-search-feedstock](https://github.com/TileDB-Inc/tiledb-vector-search-feedstock/blob/adc035b1ac6e4cbfec70e12d8b0c41fa3121c671/recipe/meta.yaml#L19) are different. The recipe does not have a separate output for the C++ library. Since the C++ library is always vendored with the Python package, there is no need to have a command line flag to help the Python package find it.

TileDB-VCF is more similar to TileDB-SOMA, and it is still using [`setup.py`](https://github.com/single-cell-data/TileDB-SOMA/blob/8261e6edac8c9cee692b06e7df15fa8f379f8b39/apis/python/setup.py). In additional to passing command line flags to `setup.py`, it also supports setting env vars so that you can use `pip install`.

I don't know if we have an example Python package that uses scikit-build-core but also maintains the feature of finding an existing shared library. @dudoslav and @ihnorton will have to advise on that aspect.

jdblischak:
Also note that the [macOS Python build](https://github.com/TileDB-Inc/TileDB-VCF/actions/runs/8664894939/job/23763038151?pr=701) works without `--libtiledbvcf` only because I put `libtiledbvcf` in `dist/lib`, which is exactly where CMake looks for it.

https://github.com/TileDB-Inc/TileDB-VCF/blob/23565d2a074722fc7297f1d6d7169126d50d3991/.github/workflows/macos.yml#L75-L79

Also, as an aside, I see that this macOS job suffers from the same SIP problems I just fixed yesterday in https://github.com/single-cell-data/TileDB-SOMA/pull/2435. That is why `DYLD_LIBRARY_PATH` is [empty](https://github.com/TileDB-Inc/TileDB-VCF/actions/runs/8664894939/job/23763038151?pr=701#step:6:16). I'll address that in a separate PR later.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
...
+    VERSION ${SKBUILD_PROJECT_VERSION}
+    LANGUAGES CXX
+)
+
+find_package(pybind11 REQUIRED)
+
+find_library(
+    TILEDBVCF_LIB
+    NAMES tiledbvcf
+    PATHS "${CMAKE_CURRENT_SOURCE_DIR}/../../dist/lib"
```
ihnorton, line 38: Not sure if this is the direct cause of the error, but this will almost certainly be `/bin` on windows. I believe if you remove `/lib` then `find_library` will search both on Windows.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
...
+        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/../../libtiledbvcf/build"
+    )
+
+    find_library(
+        TILEDBVCF_LIB
+        NAMES tiledbvcf
+        PATHS "${CMAKE_CURRENT_SOURCE_DIR}/../../dist/lib"
+    )
+endif()
+
```
ihnorton, line 58: Should probably check `TILEDBVCF_LIB-FOUND` here to make sure one of the statements succeeded.

ihnorton (commented):
🎉. One high-level comment: we should probably make the `libtiledbvcf` cmake build do a configured install that can be found with `find_package` instead of the individual `find_library` calls for libtiledbvcf. Probably not necessary to get this PR working, but if we don't do it here then @dudoslav could take a look as a backlog item.

apis/python/CMakeLists.txt
ihnorton: Oh, hmm maybe not IIUC this override https://github.com/TileDB-Inc/TileDB-VCF/blob/23565d2a074722fc7297f1d6d7169126d50d3991/libtiledbvcf/src/CMakeLists.txt#L335

gspowley:
> I think the Windows failure is due to the loss of the flag --libtiledbvcf. 

Thanks @jdblischak, this was the problem. We can now pass in the location of libtiledbvcf with the `LIBTILEDBVCF_DIR` environment variable.

> we should probably make the libtiledbvcf cmake build do a configured install that can be found with find_package instead of the individual find_library calls for libtiledbvcf.

I agree @ihnorton. This PR works for now and @dudoslav could improve it in the future to be more idiomatic.

awenocur (approved):

apis/python/pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+]
+
```
dudoslav, line 43: I see that this `pyproject.toml` is missing dependencies, however, even the `setup.py` does not have any requires (except the build_requires). Doesn't this project really depend on any other packages? (tiledb, tiledb-cloud, numpy, etc?)

apis/python/pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+[build-system]
+requires = ["scikit-build-core", "pybind11[global]"]
+build-backend = "scikit_build_core.build"
+
+[tool.scikit-build]
+build-dir = "build/{wheel_tag}"
+metadata.version.provider = "scikit_build_core.metadata.setuptools_scm"
+sdist.include = ["src/tiledbvcf/version.py"]
+sdist.cmake = true
```
dudoslav, line 56: I don't think that this is needed. This just runs CMake during `sdist` packaging process. In Vector-Search we use this to generate version files/query git status, since `.git` is not present in `sdist` package.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+cmake_minimum_required(VERSION 3.21)
+
+project(
+    ${SKBUILD_PROJECT_NAME}
```
dudoslav, line 28: I prefer not to use `SKBUILD_PROJECT_NAME` because it implies that we are going to run this `CMakeLists.txt` only in `scikit-build-core` environment. Same goes for the version. But this is just a suggestion, if you think that no-one will ever call this cmake file on its own then keep it as it is.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+# specified by the LIBTILEDBVCF_DIR environment variable.
+set(LIBTILEDBVCF_SEARCH_PATHS 
+    "${CMAKE_CURRENT_SOURCE_DIR}/../../dist/lib" 
+    "$ENV{LIBTILEDBVCF_DIR}"
+)
+
+message(STATUS "Searching for libtiledbvcf in ${LIBTILEDBVCF_SEARCH_PATHS}")
+
+# TODO: Use find_package() instead of find_library() to find libtiledbvcf.
+find_library(
```
dudoslav, line 45: Can't we use find_package? Or if this is not a different package but build by this project then add_subdirectory?

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+    src/tiledbvcf/binding/reader.cc
+    src/tiledbvcf/binding/writer.cc
+    src/tiledbvcf/binding/vcf_arrow.cc
+)
+
+target_compile_features(${VCF_TARGET_NAME} PRIVATE cxx_std_20)
+
+target_include_directories(
+    ${VCF_TARGET_NAME} PRIVATE
+    ${pybind11_INCLUDE_DIR}
```
dudoslav, line 94: This should not be needed. PYBIND headers are included automatically by pybind11_add_module.

apis/python/build-wheels.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
dudoslav, line 1: Can we use `cibuildwheel` instead of this script?

apis/python/requirements-dev.txt
```diff
@@ -0,0 +1,7 @@
+black
```
dudoslav, line 1: What is the difference between the dev requirements and requirements for the package? (This is genuine question I don't know :D )

dudoslav (commented):
Just some minor nit-picks and questions. Before merging please:

1. Create `sdist` and `wheel` packages
2. For each of these, install them and print out their version (something like `print(tilevcf.version)` if it exists)
3. Optional: Also test on Windows, because I remember issues with it in TileDB-Vector-Search

jdblischak:
The nightly builds against libtiledb 2.20 are failing with the CMake 2.29.1 error. @gspowley please rebase your PR onto main to incorporate 23565d2a074722fc7297f1d6d7169126d50d3991, (#700), which bumps the nightlies to build against libtiledb 2.21 (which had the fix backported)

jdblischak:
> this was the problem

Great! Glad this was able to fix the Windows build

>  We can now pass in the location of libtiledbvcf with the `LIBTILEDBVCF_DIR` environment variable.

@gspowley how strong is your preference for `LIBTILEDBVCF_DIR` over the existing `LIBTILEDBVCF_PATH` env var? Are you following a convention for these types of env vars? I would default to keeping the env var name the same for two primary motivations:

1. Avoid breaking any existing setups. Currently users can install tiledbvcf-py with `pip install` plus `LIBTILEDBVCF_PATH`. If the env var name is kept the same, the transition to scikit-build-core would be seamless
2. Consistency with TileDB-SOMA. The `setup.py` for tiledbsoma-py uses the env vars [`TILEDB_PATH` and `TILEDBSOMA_PATH`](https://github.com/single-cell-data/TileDB-SOMA/blob/8fd9357cb9688ea3fbf3f183cc55a7d744bb226f/apis/python/setup.py#L57-L59)

apis/python/pyproject.toml
jdblischak: This has been a long-standing frustration of mine since it makes it difficult to know what needs to be installed (eg for nightly builds or conda recipes), eg most recently in https://github.com/TileDB-Inc/TileDB-VCF/pull/669.

I think this overhaul of the Python build system is a good time to explicitly list the [required and optional dependencies](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements)

apis/python/requirements-dev.txt
jdblischak: Similar to my comment above (https://github.com/TileDB-Inc/TileDB-VCF/pull/701/files#r1565992899), I agree that this file `requirements-dev.txt` is confusing because it includes more than just development-only dependencies. Here is how I would classify them based on my current understanding:

* dev (only needed for local development and CI): black
* runtime (required to `import tiledbvcf` because they are hit by `__init__.py`): pandas, pyarrow, pyarrow-hotfix
* optional (conditionally used if installed): dask
* test (required to run the test suite): dask, pytest, tiledb

apis/python/pyproject.toml
gspowley: Agreed. Since VCF has been a conda only install so far, we've push dependency management to the conda environments. We will make the dependencies clear in `pyproject.toml`.

apis/python/pyproject.toml
gspowley: Ok, we can remove this.

apis/python/CMakeLists.txt
gspowley: I expect this `CMakeLists.txt` will only be called in the `scikit-build-core` environment. Let's keep it as.

apis/python/CMakeLists.txt
gspowley: Yes, that was @ihnorton's comment too. The `libtiledbvcf` cmake files will need some work to support a different approach. I'll add that task as a backlog item instead of blocking this PR.

apis/python/build-wheels.sh
gspowley: Good suggestion. These are VCF's first steps towards building wheels and suggestions for standard processes are welcome.

apis/python/requirements-dev.txt
gspowley: The `requirements-dev.txt` was only there for building a local dev environment. As mentioned above, we'll make the dependencies clear in a more standard way.

jdblischak:
I tested this PR in my fork of tiledb-vcf-feedstock.

* [branch](https://github.com/jdblischak/tiledb-vcf-feedstock/tree/scikit-build-core)
* [changes](https://github.com/TileDB-Inc/tiledb-vcf-feedstock/compare/master...jdblischak:tiledb-vcf-feedstock:scikit-build-core)
* [CI](https://dev.azure.com/jdblischak/feedstock-builds/_build/results?buildId=742&view=results)

Note that the failing win-64 build is a known issue due to linking to libfmt (discussed in https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/116). The error messages are the same as in the failing CI for those most recently merged commit, https://github.com/TileDB-Inc/tiledb-vcf-feedstock/commit/761620c8be91d46be834f2ff4661a40325212ad7

gspowley:
> how strong is your preference for LIBTILEDBVCF_DIR over the existing LIBTILEDBVCF_PATH

@jdblischak not very strong. I was pattern matching the environment variables for finding `tiledb`, `htslib`, `spdlog`, `catch`, and `cli11`. I value not breaking things over matching names, so we can go back to `LIBTILEDBVCF_PATH`.

apis/python/CMakeLists.txt
gspowley: Made windows happy with proper path manipulation, which could move to an improved `libtiledbvcf` cmake structure.

jdblischak:
I tested the most recent changes in the feedstock. The linux and osx builds are now failing with the following configuration error that has something to do with aws-c-auth. Note that I replaced the long environment path with `$CONDA_PREFIX` to improve readability:

https://dev.azure.com/jdblischak/feedstock-builds/_build/results?buildId=744&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3

```sh
  -- Found PythonInterp: $CONDA_PREFIX/bin/python (found suitable version "3.10.14", minimum required is "3.6")
  -- Found PythonLibs: $CONDA_PREFIX/lib/libpython3.10.so
  -- Performing Test HAS_FLTO
  -- Performing Test HAS_FLTO - Success
  -- Found pybind11: $CONDA_PREFIX/lib/python3.10/site-packages/pybind11/include (found version "2.12.0")
  -- Searching for libtiledbvcf in /home/conda/feedstock_root/build_artifacts/tiledb-vcf_1713276872393/work/apis/python/../../dist/lib;
  -- Found libtiledbvcf: $CONDA_PREFIX/lib/libtiledbvcf.so
  CMake Error at CMakeLists.txt:122 (install):
    install FILES given directory
    "$CONDA_PREFIX/lib/aws-c-auth"
    to install.


  -- Configuring incomplete, errors occurred!

  *** CMake configuration failed
```

Also, today I noticed some CMake warnings. These were there previously, so they aren't fatal, but they might be worth addressing:

```
  Building wheel for tiledbvcf (pyproject.toml): started
  Running command Building wheel for tiledbvcf (pyproject.toml)
  *** scikit-build-core 0.8.2 using CMake 3.29.2 (wheel)
  *** Configuring CMake...
  2024-04-16 14:26:58,194 - scikit_build_core - WARNING - Unsupported CMAKE_ARGS ignored: -DCMAKE_BUILD_TYPE=Release
  2024-04-16 14:26:58,194 - scikit_build_core - WARNING - Unsupported CMAKE_ARGS ignored: -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX
  2024-04-16 14:26:58,194 - scikit_build_core - WARNING - libdir/ldlibrary: $CONDA_PREFIX/lib/libpython3.10.a is not a real file!
  2024-04-16 14:26:58,194 - scikit_build_core - WARNING - Can't find a Python library, got libdir=$CONDA_PREFIX/lib, ldlibrary=libpython3.10.a, multiarch=x86_64-linux-gnu, masd=None
  2024-04-16 14:26:58,348 - scikit_build_core - WARNING - Unsupported CMAKE_ARGS ignored: -DCMAKE_BUILD_TYPE=Release
  2024-04-16 14:26:58,348 - scikit_build_core - WARNING - Unsupported CMAKE_ARGS ignored: -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX
  loading initial cache file build/cp310-cp310-manylinux_2_17_x86_64/CMakeInit.txt
```

gspowley:
> The linux and osx builds are now failing with the following configuration error

Thanks for the heads up. I'll follow up with a PR to fix this later today.

PR #702, opened 2024-04-17
Fix python cmake issue

gspowley:
Fix an issue introduced in #701 by providing a list of shared libs to install explicitly.


awenocur (approved):

jdblischak:
Testing now in https://github.com/jdblischak/tiledb-vcf-feedstock/commit/06b18423b7526cb2ac6d1cf702d350ce723c7fd3

gspowley:
Thanks @jdblischak. Based on your test, I see we need to be explicit about the shared libraries installed. The latest commit should install only the required libraries.

jdblischak:
@gspowley Thanks! I manually triggered another CI run to test your latest changes

https://dev.azure.com/jdblischak/feedstock-builds/_build/results?buildId=748&view=results

jdblischak (approved):
Feedstock build passed

PR #703, opened 2024-04-18
Use HTSlib 1.19.0 on Windows.

teo-tsirpanis:
This is the latest available version in https://github.com/TileDB-Inc/m2w64-htslib-build; there is also version 1.19.1 which is used on non-Windows.

Fixes compile errors in places that use `VCF_INS` and `VCF_DEL`, including here: https://github.com/TileDB-Inc/TileDB-VCF/blob/964be5c736f89675d9308d4eecd7e07294319d1d/libtiledbvcf/src/stats/sample_stats.cc#L238


gspowley (approved):
Thanks @teo-tsirpanis!

PR #708, opened 2024-04-26
Sample stats improvements

gspowley:
1. Add option to disable sample stats at dataset creation time.
2. Support deleting a sample from the array.
3. Handle missing DP and GQ properly.
4. Tests for all of the above.


awenocur (approved):

leipzig:
this is fine as long as we don't expect the same thing to happen to other fields

leipzig (approved):

PR #725, opened 2024-05-28
Write additional VCF variant stats during ingestion

gspowley:
Write additional stats needed to calculate `variant_qc` to the `variant_stats` array during ingestion.


gspowley:
Closing since these changes are no longer needed.

PR #734, opened 2024-06-24
Hotfix: pin numpy < 2

jdblischak:
Both the nightlies in this repo (#730) and my centralized nightlies (https://github.com/jdblischak/centralized-tiledb-nightlies/issues/12#issuecomment-2176240410) have been failing for a week due to the release of numpy 2. PRs #731 and #732 have the numpy<2 pin, but they are still undergoing review. To fix the nightly builds, I cherry-picked @awenocur's commit to pin numpy<2.

PR #719, opened 2024-05-03
Sample stats improvements

gspowley:
This PR modifies the sample stats to match the results of Hail's `sample_qc` and adds a `tiledbvcf.sample_qc` aggregation function implemented in c++.

PR #579, opened 2023-09-19
Nightly build for macOS

jdblischak:
The nightly macOS build fails right at the end of the installation of the Python package ([link to failed build on my fork](https://github.com/jdblischak/TileDB-VCF/actions/runs/6237936964/job/16932652886#step:13:114)). I don't understand why it is searching PyPI for a locally built package, and I wasn't able to find any obvious solution when searching the error message. Has anyone observed this error before?

```sh
Processing tiledbvcf-0.26.1.dev2-py3.9-macosx-10.9-x86_64.egg
creating /Users/runner/micromamba/envs/py4vcf/lib/python3.9/site-packages/tiledbvcf-0.26.1.dev2-py3.9-macosx-10.9-x86_64.egg
Extracting tiledbvcf-0.26.1.dev2-py3.9-macosx-10.9-x86_64.egg to /Users/runner/micromamba/envs/py4vcf/lib/python3.9/site-packages
Adding tiledbvcf 0.26.1.dev2 to easy-install.pth file
Installed /Users/runner/micromamba/envs/py4vcf/lib/python3.9/site-packages/tiledbvcf-0.26.1.dev2-py3.9-macosx-10.9-x86_64.egg
Processing dependencies for tiledbvcf==0.26.1.dev2
Searching for tiledbvcf==0.26.1.dev2
Reading https://pypi.org/simple/tiledbvcf/
Couldn't find index page for 'tiledbvcf' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.org/simple/
No local packages or working download links found for tiledbvcf==0.26.1.dev2
error: Could not find suitable distribution for Requirement.parse('tiledbvcf==0.26.1.dev2')
```

PR #740, opened 2024-07-01
Build with numpy 2

jdblischak:
Follow-up to #734 

What is preventing us from building TileDB-VCF with numpy 2? Help wanted

### DIFF
--- a/apis/python/conda-env.yml
+++ b/apis/python/conda-env.yml
@@ -2,7 +2,7 @@ name: tiledbvcf-py
 channels:
   - conda-forge
 dependencies:
-  - numpy<2
+  - numpy
   - python<3.12 # based on available tiledb-py wheels
   - pybind11
   - pyarrow>=14.0.2 # for pyarrow security fix
--- a/apis/python/pyproject.toml
+++ b/apis/python/pyproject.toml
@@ -40,7 +40,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 
-dependencies = ["pandas", "pyarrow", "pyarrow-hotfix", "numpy<2"]
+dependencies = ["pandas", "pyarrow", "pyarrow-hotfix", "numpy"]
 
 [project.optional-dependencies]
 test = ["dask[distributed]", "pytest", "tiledb"]
--- a/ci/gha-win-env.yml
+++ b/ci/gha-win-env.yml
@@ -11,7 +11,7 @@ dependencies:
   - tiledb=2.15
   - vs2019_win-64
   # build tiledbvcf-py
-  - numpy<2
+  - numpy
   - pandas<2.0
   - pyarrow=9.0
   - pyarrow-hotfix


jdblischak:
[sc-50106]

jdblischak:
It appears to fail whenever tiledbvcf-py is imported on Linux or macOS. The import passed on Windows. The passing Linux and macOS builds don't run `import tiledbvcf`, and thus avoid the error.


```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "<string>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/tiledbvcf/__init__.py", line 2, in <module>
    import pyarrow
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pyarrow/__init__.py", line 65, in <module>
    import pyarrow.lib as _lib
AttributeError: _ARRAY_API not found
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/tiledbvcf/__init__.py", line 2, in <module>
    import pyarrow
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pyarrow/__init__.py", line 65, in <module>
    import pyarrow.lib as _lib
  File "pyarrow/lib.pyx", line 37, in init pyarrow.lib
ImportError: numpy.core.multiarray failed to import
```

gspowley:
It looks like this is related to the `pyarrow==11` pin.  I reproduced the error locally (on Linux) by pinning `pyarrow==11` and avoided the error by removing the pin.

gspowley:
An overly aggressive unpinning experiment in #741 failed on windows and macos. I believe the macos failure is known and why we're pinning `pyarrow` in the first place.

For reference, the build and pytests succeeded on a local macos 14.5 arm64 with these versions:
```
numpy              2.0.0
pyarrow            16.1.0
```

jdblischak:
@gspowley thanks for investigating! So at least for VCF, it looks like the blocker for numpy 2 is this issue with pyarrow

jdblischak:
Cross-referencing numpy<2 PRs:

* SOMA has pinned numpy<2 https://github.com/single-cell-data/TileDB-SOMA/pull/2745 (https://github.com/single-cell-data/TileDB-SOMA/issues/2536)
* Vector Search has pinned numpy<2 https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/419

gspowley:
The `pyarrow` pin for the macos nightly CI was introduced in #579. The debugging in #719 included attempts to remove the pyarrow pins, which failed for macos, so the pins were maintained.

@jdblischak and @awenocur may have more context on the original macos failures.

jdblischak:
xref: https://github.com/TileDB-Inc/TileDB-VCF/pull/762, https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/138

gspowley:
Addressed in #762.

PR #742, opened 2024-07-02
Setuptools SCM version file

dudoslav:
I believe that the correct way to set version file is to use
`write_to`
instead of
`version_file`

https://scikit-build-core.readthedocs.io/en/stable/configuration.html#dynamic-metadata

* Note that I found this out when building with Conda recipe for Cloud-Images

1719913747_grim

### DIFF
--- a/apis/python/pyproject.toml
+++ b/apis/python/pyproject.toml
@@ -66,4 +66,4 @@ editable.mode = "inplace"
 
 [tool.setuptools_scm]
 root = "../.."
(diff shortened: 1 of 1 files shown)


dudoslav:
My bad, actually `write_to` is deprecated:

```
(deprecated) legacy option to create a version file relative to the scm root it's broken for usage from a sdist and fixing it would be a fatal breaking change, use version_file instead.
```

Closing this PR.

PR #757, opened 2024-08-13
reduce latency creating datasets

awenocur:


gspowley:
Closing since this did not reduce the latency much.

PR #759, opened 2024-08-17
add Python API for sample deletion

awenocur:
Add a Python API to delete samples.


libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -938,7 +938,9 @@ void TileDBVCFDataset::delete_samples(
     const std::vector<std::string>& sample_names,
     const std::vector<std::string>& tiledb_config) {
   // Open dataset in read mode, required before calling `sample_exists`.
-  open(uri);
+  if (!open_) {
+    open(uri);
```
gspowley, line 942: The `tiledb_config` needs to be passed to `open`.

gspowley (changes requested):
Looks good. One change requested and please add a description to the PR. Thanks!

gspowley (approved):

PR #761, opened 2024-08-20
Handle legacy stats arrays

gspowley:
The computation of variant statistics was improved in [0.33.0](https://github.com/TileDB-Inc/TileDB-VCF/releases/tag/0.33.0). This PR handles writing to the `sample_stats` and `variant_stats` arrays created before version 0.33.0.

### DIFF
--- a/libtiledbvcf/src/stats/variant_stats.cc
+++ b/libtiledbvcf/src/stats/variant_stats.cc
@@ -38,6 +38,9 @@ namespace tiledb::vcf {
 int32_t VariantStats::max_length_ = 0;
 
 uint32_t VariantStats::array_version_ = VariantStats::VARIANT_STATS_MIN_VERSION;
+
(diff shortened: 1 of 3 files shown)


libtiledbvcf/src/stats/variant_stats.cc
```diff
@@ -213,6 +216,10 @@ void VariantStats::init(std::shared_ptr<Context> ctx, const Group& group) {
     throw std::runtime_error(
         "encountered variant stats array version out of range while writing");
 
+  // Check for presence of "an" attribute
+  auto schema = fetch_version.schema();
+  an_present_ = schema.has_attribute("an");
```
awenocur, line 221: Why is this this checking the presence of a field, rather than using `array_version_`?

awenocur (approved):

spencerseale (approved):
Thank you!

PR #763, opened 2024-08-28
Draft: Add initial build wheels CI/CD pipeline

dudoslav:
This PR is being tested in my personal fork.

PR #762, opened 2024-08-26
Build with NumPy 2

gspowley:
Changes in this PR
* Remove `numpy` and `pyarrow` pins
* Remove python 3.8 (not supported by numpy 2) and add python 3.12
* Add changes for windows build in conda feedstock
* Remove unused Dockerfile and env file

See https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/138 for conda feedstock experiments using this branch.

PR #766, opened 2024-09-03
Build TileDB-VCF-Py on Windows without conda

jdblischak:
**Goal:** Run Windows CI job without using conda

**Motivation:** We haven't been able to build the TileDB-VCF conda binary for months due to problems linking to fmt/spdlog from conda-forge. Ideally these sort of conda-specific problems should be handled in the feedstock repo, not here. It wasn't a pressing concern until recently when @gspowley started work to migrate to numpy 2, which included updating pyarrow (https://github.com/TileDB-Inc/TileDB-VCF/pull/762#issuecomment-2316301233). Now the Windows CI build is failing due to similar fmt/spdlog linking errors.

**Status:** I can build libtiledbvcf and tiledbvcf-py on Windows without conda. However, I can't `import tiledbvcf` because it can't find the libtiledbvcf DLL (strangely, since it is installed in the Python package).

**More details:**

* Ideally, since we are using scikit-build-core, the build would be as straightforward as `python -m pip install -v apis\python[test,dev]`. However, this failed because the following wouldn't execute on Windows:

https://github.com/TileDB-Inc/TileDB-VCF/blob/b9fd4386a2bcfb07e7eb59c7b34ee736844786a3/apis/python/CMakeLists.txt#L55-L61

* The Python CMake expects TileDB to be in `apis/python/dist_links/libtiledbvcf/build/externals/install`. This caused me headaches when I tried building libtiledbvcf from the root of the directory. I tried a few different ways to build from the symlinked subdirectory `apis/python/dist_links/libtiledbvcf`, but symlinks are different on Windows, so I abandoned this

https://github.com/TileDB-Inc/TileDB-VCF/blob/b9fd4386a2bcfb07e7eb59c7b34ee736844786a3/apis/python/CMakeLists.txt#L104-L107

* I finally got everything to build by setting `TileDB_DIR=.\libtiledbvcf\build\externals\install\lib\cmake\TileDB\`, so that the Python CMake routine could fine TileDB

```
  *** Installing project into wheel...
  -- Installing: C:\Users\RUNNER~1\AppData\Local\Temp\tmpc7c23i0p\wheel\platlib/tiledbvcf/libtiledbvcf.cp311-win_amd64.pyd
  -- Installing: C:\Users\RUNNER~1\AppData\Local\Temp\tmpc7c23i0p\wheel\platlib/tiledbvcf/tiledb.lib
  -- Installing: C:\Users\RUNNER~1\AppData\Local\Temp\tmpc7c23i0p\wheel\platlib/tiledbvcf/tiledbvcf.lib
  -- Installing: C:\Users\RUNNER~1\AppData\Local\Temp\tmpc7c23i0p\wheel\platlib/tiledbvcf/hts-3.lib
  -- Installing: C:/Users/RUNNER~1/AppData/Local/Temp/tmpc7c23i0p/wheel/scripts/tiledbvcf.dll
  -- Installing: C:/Users/RUNNER~1/AppData/Local/Temp/tmpc7c23i0p/wheel/scripts/tiledbvcf.exe
  -- Installing: C:/Users/RUNNER~1/AppData/Local/Temp/tmpc7c23i0p/wheel/scripts/tiledbvcf4test.dll
```

* However, I can't run the tests because I can't even import tiledbvcf-py. Unclear what I can do about this. Maybe the problem is that only `tiledb.lib` is copied to the Python package and not `tiledb.dll`

```sh
python -c "import tiledbvcf; print(tiledbvcf.version)"
## Traceback (most recent call last):
##   File "<string>", line 1, in <module>
##   File "C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\tiledbvcf\__init__.py", line 9, in <module>
##     import tiledbvcf.libtiledbvcf
## ImportError: DLL load failed while importing libtiledbvcf: The specified module could not be found.
```

https://github.com/TileDB-Inc/TileDB-VCF/blob/b9fd4386a2bcfb07e7eb59c7b34ee736844786a3/apis/python/src/tiledbvcf/__init__.py#L9

* Note that I fixed the htslib download by updating the sha, which I learned from the PR https://github.com/TileDB-Inc/TileDB-VCF/pull/763 from @dudoslav

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,31 +18,27 @@ on:
       - 'ci/gha-win-env.yml'
       - 'libtiledbvcf/**'
   workflow_dispatch:
-defaults:
-  run:
-    shell: cmd /C CALL {0}
 jobs:
   build:
     runs-on: windows-2022
     steps:
       - uses: actions/checkout@v4
-      - name: Install conda env
-        uses: mamba-org/setup-micromamba@v1
         with:
-          environment-file: ci/gha-win-env.yml
-          cache-environment: true
-          init-shell: cmd.exe
-          condarc: |
-            channels:
-              - conda-forge
-              - tiledb
-            channel_priority: strict
-      - name: Build libtiledbvcf
-        run: cmd /C CALL ci\build-libtiledbvcf.bat
-      - name: libtiledbvcf version
-        run: tiledbvcf.exe version
-      - name: Build tiledbvcf-py
-        run: cmd /C CALL ci\build-tiledbvcf-py.bat
+          fetch-depth: 0 # fetch everything for python setuptools_scm
+      - name: Build and install libtiledbvcf
+        run: |
+          cmake -S libtiledbvcf -B libtiledbvcf\build
+          cmake --build libtiledbvcf\build -j2
+          cmake --build libtiledbvcf\build -j2 --target install-libtiledbvcf
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install tiledbvcf-py
+        run: |
+          set TileDB_DIR=.\libtiledbvcf\build\externals\install\lib\cmake\TileDB\
+          cd apis\python
+          python -m pip install -v .[test,dev]
       - name: tiledbvcf-py version
         run: python -c "import tiledbvcf; print(tiledbvcf.version)"
       - name: Test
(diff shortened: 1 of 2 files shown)


dudoslav:
@jdblischak I also put some effort in fixing the wheel production here: https://github.com/TileDB-Inc/TileDB-VCF/pull/763

In my case I realized that some parts of VCF build logic needs to be rewritten.

jdblischak:
I tried a few things on a separate branch, but still failing.

* Tried an editable install with `-e` since sometimes this can help with resolving finding shared libraries. This didn't help.
* Using the same editable install, I skipped the simple import test and ran `pytest`, with the hope that this scenario might be able to find the shared libraries. Unfortunately not. Still failed with the same import error

jdblischak:
Reminder of our options for fixing the Windows CI builds:

* Fix the libfmt linking issues when building against more recent versions of libtiledb that link to conda-installed versions of fmt/spdlog (https://github.com/TileDB-Inc/TileDB-VCF/pull/762#issuecomment-2430043453). This would allow us to continue to use the conda-based build setup we are currently using
* Fix the non-conda Windows build that I have attempted in this PR. I can build everything, but I can't import tiledbvcf-py (and thus also can't run its test suite). Potentially some of the updates to the Windows build in #763 might be able to help with this

PR #770, opened 2024-09-11
Add Dataset delete to the Python API

gspowley:
Add `Dataset.delete` to the Python API, which deletes TileDB-VCF datasets stored on the local file system, remote object store, or registered on TileDB Cloud.

For example:
```python
ds_uri = ...
tiledbvcf.Dataset.delete(ds_uri)
```


spencerseale (approved):
Nice add!

PR #781, opened 2024-10-17
Build System Rework

dudoslav:
This PR reworks the build system in order to properly build and package TileDB-VCF for all platforms (Linux, MacOS) and all package managers (PyPI and Conda).

The main goals of this PR:
- Divide build into TileDB-VCF and TileDB-VCF-Py
- TileDB-VCF needs to build as shared object and as a static library
- TileDB-VCF needs to internally link with shared libraries or static libraries (fmt, spdlog, cli11)
- TileDB-VCF-Py needs to properly package for PyPI, currently statically linked wheel
- TileDB-VCF-Py conda packaging logic, only build TileDB-VCF-Py and dynamically link to TileDB-VCF (that also uses shared libs)

Windows build and package does not work yet.

### DIFF
--- a/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
+++ b/libtiledbvcf/src/dataset/tiledbvcfdataset.cc
@@ -165,6 +165,9 @@ std::vector<std::string> TileDBVCFDataset::get_vcf_attributes(std::string uri) {
 
   // Read VCF header into string for parsing
   bcf_hdr_t* hdr = VCFUtils::hdr_read_header(uri);
+  if (hdr == nullptr) {
+    throw std::runtime_error("Error reading VCF header from: " + uri);
+  }
   std::string hdrstr = VCFUtils::hdr_to_string(hdr);
   const char* p = hdrstr.c_str();
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         os: [macos-13, ubuntu-24.04]
         branches:
-          - {libtiledb: release-2.26, tiledb-py: 0.32.0}
+          - {libtiledb: release-2.27, tiledb-py: 0.33.3}
           - {libtiledb: main, tiledb-py: main}
     env:
       MACOSX_DEPLOYMENT_TARGET: 10.15
@@ -50,7 +50,9 @@ jobs:
         run: bash TileDB-VCF/ci/nightly/build-libtiledb.sh
       - name: Setup to build htslib from source (Linux)
         if: runner.os == 'Linux'
-        run: sudo apt-get install --yes automake autoconf libbz2-dev liblzma-dev
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes automake autoconf libbz2-dev liblzma-dev
       - name: Setup to build htslib from source (macOS)
         if: runner.os == 'macOS'
         run: brew install autoconf automake
--- a/libtiledbvcf/src/vcf/vcf_utils.cc
+++ b/libtiledbvcf/src/vcf/vcf_utils.cc
@@ -62,6 +62,12 @@ uint32_t VCFUtils::get_end_pos(
 
 bcf_hdr_t* VCFUtils::hdr_read_header(const std::string& path) {
   auto fh = vcf_open(path.c_str(), "r");
+
+  int retries = 3;
+  while (!fh && retries--) {
+    fh = vcf_open(path.c_str(), "r");
+  }
+
   if (!fh)
     return nullptr;
   auto hdr = bcf_hdr_read(fh);


libtiledbvcf/src/dataset/tiledbvcfdataset.cc
```diff
@@ -270,7 +270,8 @@ void TileDBVCFDataset::create(const CreationParams& params) {
 
   // Log the group structure
   group.open(TILEDB_READ);
-  LOG_DEBUG("TileDB Groups: \n{}", group.dump(true));
+  // NOTE: On windows group.dump produces memory access violation
+  // LOG_DEBUG("TileDB Groups: \n{}", group.dump(true));
```
teo-tsirpanis, line 274: I would have entirely removed it, or condition it on debug-level logs being enabled (we don't have an `is_enabled` global function). Dumping groups is quite expensive as it involves several group open and I/O operations.

libtiledbvcf/src/stats/column_buffer.cc
```diff
@@ -178,7 +178,7 @@ size_t ColumnBuffer::update_size(const Query& query) {
     if (offsets_.capacity() < num_offsets + 1) {
       offsets_.reserve(num_offsets + 1);
     }
-    offsets_[num_offsets] = num_elements;
+    *(offsets_.data() + num_offsets) = num_elements;
```
teo-tsirpanis, line 181: Why?

dudoslav: MSVC had an issue with accessing offsets_ vector out of bounds, which it actually does since it uses `capacity/reserve` instead of `size/resize`. I tried to change it in order to remove this bound check. I was getting array access out of bounds.

libtiledbvcf/src/CMakeLists.txt
```diff
@@ -291,52 +245,85 @@ set_target_properties(tiledbvcf
...
+#          RUNTIME DESTINATION tiledbvcf
+#  )
+#endif()
+
+if(TILEDBVCF_INSTALL_TILEDB)
+  install(
+          IMPORTED_RUNTIME_ARTIFACTS
+          TileDB::tiledb_shared
+          RUNTIME DESTINATION tiledbvcf
+          LIBRARY DESTINATION tiledbvcf/lib
```
teo-tsirpanis, line 285: We should be using `bin` and `lib`.

ihnorton: `CMAKE_INSTALL_{BINDIR,LIBDIR}` for full windows compatibility? https://github.com/TileDB-Inc/TileDB/blob/738f238d99776e03d2eb36ddf0cd55254f9ba32f/tiledb/CMakeLists.txt#L852

dudoslav: Definitely, after the SEGFAULT is fixed I'll clean up the changes to be more organized and use CMAKE_* variables

dudoslav: This is kind of an issue, since `CMAKE_INSTALL_LIBDIR` points to `site-packages/lib` which is not correct for our case. Looking at other projects they also use just folder names: https://github.com/ssciwr/clang-format-wheel/blob/dc29f5e2a074dee8e862eb29fd0de1d84f41707a/CMakeLists.txt#L43

I don't think that we need to use `GNUInstallDirs` CMake module. However, perhaps this will come into play when I'll be testing Conda packaging.

apis/python/tests/test_tiledbvcf.py
```diff
@@ -930,32 +930,33 @@ def test_sample_and_region_partitioned_read():
...
-        ),
-    )
-
-    # total number of exported records
-    assert df.shape[0] == 1172081
-
-    # number of unique exported records
-    record_index = ["sample_name", "contig", "pos_start"]
-    assert df[record_index].drop_duplicates().shape[0] == 1168430
+# TODO: DO NOT MERGE THIS COMMENTED OUT
```
ihnorton, line 933: Flagging this to address before merge

ihnorton (changes requested):

apis/python/CMakeLists.txt
```diff
@@ -30,97 +30,41 @@ project(
...
-        COMMAND bash -c "\
-            ${CMAKE_COMMAND} .. && \
-            ${CMAKE_COMMAND} --build . && \
-            ${CMAKE_COMMAND} --build . --target install-libtiledbvcf"
-        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/dist_links/libtiledbvcf/build"
-    )
+if (NOT TARGET tiledbvcf)
+    message(STATUS "DEBUG: tiledbvcf_DIR: ${tiledbvcf_DIR}")
+    # If tiledbvcf is not present it means that this script was not included as a
+    # subdirectory of the tiledbvcf library, thus we need to find it
```
teo-tsirpanis, line 36: Why do we need to support that? There is already the capability in this PR to build libtiledbvcf and its python API from the same buildtree. If we do need to support that we will have to export the target with CMake config files.

dudoslav: Because of Conda feedstock logic. VCF should be divided into two packages, VCF and VCF-Py where VCF-Py should not copy any of its shared libraries into the system if they are present as Conda packages, so it needs to build VCF-Py with VCF from the system.

gspowley (changes requested):
Thank you for working on this @dudoslav. A couple of high level comments. 

Please keep the `Makefile` in the root of the repo, which a convenience tool for developers as opposed to part of the build system.

Releasing the build system changes with pypi upload is more important than enabling windows support in this PR.

PR #784, opened 2024-10-28
type return incorrectly inst an object

spencerseale:
The type hint for the return instantiates a dataframe rather than just hints. This doesn't really cause an issue, but does cause that dataframe to be created whenever importing `tiledbvcf` do to this module being imported in top-level `__init__.py`


Shelnutt2:
Please add a description to the PR.

Shelnutt2 (approved):

PR #787, opened 2024-11-04
Change VLA code by using std::vector

dudoslav:
In order to fix an issue with newer compilers we have to remove any usage of VLA from VCF code. I am not sure how this changes the performance of the code, but for now I decided to use `std::vector`. If we could bound the size of the array we could use `std::array` and have the memory allocated on stack.

https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/141#issuecomment-2450369283


gspowley (approved):
LGTM, thanks @dudoslav!

I expect the performance overhead will be low because the number of elements allocated will be very small.

PR #783, opened 2024-10-21
Migrate to macos-13 runner image

jdblischak:
macos-12 runner image is deprecated

xref: https://github.com/actions/runner-images/issues/10721, https://github.com/TileDB-Inc/TileDB/pull/5350

PR #792, opened 2024-12-04
More robust non-empty buffer check during ingestion

gspowley:
During VCF ingestion, explicitly check that the query buffers are non-empty using `utils::query_buffers_set` before clearing the buffers.


gspowley:
This PR needs #783.

leipzig (approved):

spencerseale (approved):

PR #822, opened 2025-07-17
Updated HTSlib to version 1.22.1

alancleary:
This includes changing the CMake External Project URL hash for HTSlib to use a SHA256 checksum because GitHub now auto-generates SHA256 checksums for HTSlib release artifacts.


alancleary:
Manually ran the nightly on this branch and those pass as well. Merging

PR #824, opened 2025-07-21
Revert "Updated spdlog to version 1.15.3"

alancleary:
Reverts TileDB-Inc/TileDB-VCF#82 because this change is causing the [nightly build on Ubuntu 24.04 to fail](https://github.com/TileDB-Inc/TileDB-VCF/issues/823). Will revisit when a release is not imminent.


jdblischak (approved):

jdblischak:
I manually triggered a nightly build from this PR branch: https://github.com/TileDB-Inc/TileDB-VCF/actions/runs/16422969387

PR #828, opened 2025-08-04
[WIP] Add single variant to existing sample

alancleary:
@simonjwhite Would like the ability to add variants one at a time to support liftover. This PR adds that functionality.

Currently the feature is only available via the `add-variant` command in the CLI, which accepts a single variant as JSON. For example:
```bash
tiledbvcf add-variant -u ./1000-genomes-dragen-v3.7.6-dataset --json '{"chrom":"chr1","pos":10,"ref":"A","alt":"T","sample":"HG03914","id":"rs123","qual":30.0,"info":{"DP":45,"AF":0.3},"format":{"GT":"0/1","DP":15,"GQ":25},"filter":["PASS"]}'
```
Notes:
* Only one variant can be added at a time
* The sample a variant is being added to must already exist in the dataset
* This functionality is currently implemented as a wrapper for ingestion, i.e. it generates temporary `.vcf.gz` and `.tbi` files and then ingests them
  * The ingestion can be configured through the `add-variant` command via pass-through flags
* Temporary files are written to `/dev/shm/` for performance
  * This can be configured via the `--tmp-dir` flag

I will add tests and proceed with the Python API after getting feedback on the current implementation.

### DIFF
--- a/libtiledbvcf/src/cli/tiledbvcf.cc
+++ b/libtiledbvcf/src/cli/tiledbvcf.cc
@@ -106,6 +106,34 @@ void do_store(const IngestionParams& args, const CLI::App& cmd) {
   LOG_TRACE("Finished store command.");
 }
 
+/** Store/ingest a single variant. */
+void do_variant(const IngestionParams& args, const CLI::App& cmd) {
+  if (args.variant_json.size() == 0) {
+    std::cerr
+        << "ERROR: RequiredError: --json is required\n";
+    throw CLI::CallForHelp();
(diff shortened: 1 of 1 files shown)


spencerseale (commented):
Are you adding the python implementation to this PR? Or are we holding off for now?

alancleary:
> Are you adding the python implementation to this PR? Or are we holding off for now?

Holding off for now. We're considering making the implementation/API more transactional to reduce fragmentation.

PR #829, opened 2025-08-06
Replace strlen(...) in Buffer class with direct computation

alancleary:
strlen(...) can take up to O(n) time, depending on the implementation, whereas the direct computation takes O(1) time and uses data that is already present in the class.


alancleary:
Wrong solution here. Will open an new PR in a fresh branch.

PR #832, opened 2025-08-13
(WIP) Use offset buffer for get variable length buffer data

jp-dark:
Remove `strlen` and replace it with direct call to offset buffers.

### DIFF
--- a/libtiledbvcf/src/utils/buffer.cc
+++ b/libtiledbvcf/src/utils/buffer.cc
@@ -188,8 +188,8 @@ std::string_view Buffer::value(uint64_t element_index) const {
   assert(data_);
 
   uint64_t start = offsets_[element_index];
-  uint64_t len = strlen(data_ + start);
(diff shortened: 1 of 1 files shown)


libtiledbvcf/src/utils/buffer.cc
```diff
@@ -188,17 +188,19 @@ std::string_view Buffer::value(uint64_t element_index) const {
...
-  return std::string_view(data_ + start, len);
+  uint64_t length = offsets_[element_index + 1] - start;
+  return std::string_view(data_ + start, length);
 }
 
 std::vector<std::string_view> Buffer::data() const {
   assert(!offsets_.empty());
   assert(data_);
 
   std::vector<std::string_view> vec(offset_nelts_);
```
jp-dark, line 199: This length is longer than the number of elements (n offsets corresponds to n-1 elements).

PR #838, opened 2025-09-02
Update Docker commands since ENTRYPOINT was removed

jdblischak:
Follow-up to #816. With the `ENTRYPOINT` removed, the Docker commands in the documentation and `Makefile` need to specify `tiledbvcf`, which is on the `PATH`.


alancleary (approved):
Looks good. Thanks for the follow up!

PR #858, opened 2025-11-12
Update TileDB core to 2.29.2

alancleary:
Note that the macos-15-libtiledb-main CI will fail on this PR. The [issue](https://github.com/TileDB-Inc/TileDB-VCF/issues/855#issuecomment-3498811588) comes from a recently merged PR in TileDB Core that causes many of its dependencies to be updated. The problematic PR was not included in TileDB Core release 2.29.2. As such, the macos-15-libtiledb-release-2.29 CI will succeed for this PR, meaning it is safe to proceed with the release.


alancleary:
As expected, the [macos-15-libtiledb-main CI job failed and the macos-15-libtiledb-release-2.29 job succeeded](https://github.com/TileDB-Inc/TileDB-VCF/actions/runs/19313042227). Proceeding with [VCF release 0.39.2](https://github.com/TileDB-Inc/TileDB-VCF/pull/859).

PR #874, opened 2026-03-12
VCF-103: Buffer global order writes to reduce the number of batches submitted

alancleary:
This PR implements a new parallelization strategy for VCF ingestion. Now, instead of partitioning regions into slices and parsing them in parallel, each chromosome is treated as a single region and VCF files are partitioned into groups that are parsed in parallel using multiple workers. These workers _stream_ records into an algorithm that merges the records into a single global-order buffer, instead of many small buffers like the previous algorithm did. The result is that writes are much larger and happen less frequently.

Noteworthy points:
* In testing, this approach sends ~1/2 as many requests to S3
* In testing, this approach uses ~1/2 as much memory
* This approach is now the default that runs via the CLI and Python API; the previous algorithm is still available via the `--legacy` flag and `legacy` parameter, respectively
* Resume functionality currently isn't supported (use legacy version if necessary)
* Currently this approach only supports V4 datasets (use legacy version if necessary)
* As of the time this PR was opened, the new approach runs **slower** than the previous algorithm. This is due to stats computations affecting the hotpath and should be mitigated before merging.

More generally, computing stats is quite slow and the lack of configuration of when and how they write buffered data puts a ceiling on performance. These issues are out of scope for this PR (see VCF-62), however, it's worth noting that once they are addressed the parallelization scheme of this PR can be extended to parse multiple chromosomes in parallel, which will result in higher thread utilization and, therefore, higher throughput.


libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -581,7 +591,7 @@ void AlleleCountReader::read_from_allele_count(
   filter_offsets[0] = 0;
   gt_offsets[0] = 0;
   size_t i = 0;
-  for (std::pair<AlleleCountKey, int32_t> ac : AlleleCountGroupBy) {
+  for (std::pair<AlleleCountKey, int32_t> ac : alleleCountGroupBy) {
```
XanthosXanthopoulos, line 594: Maybe use structured binding and avoid copy while iterating

libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -525,13 +535,13 @@ void AlleleCount::update_results() {
 std::tuple<size_t, size_t, size_t, size_t, size_t>
 AlleleCountReader::allele_count_buffer_sizes() {
   size_t ref = 0, alt = 0, filter = 0, gt = 0;
-  for (std::pair<AlleleCountKey, int32_t> ac : AlleleCountGroupBy) {
+  for (std::pair<AlleleCountKey, int32_t> ac : alleleCountGroupBy) {
```
XanthosXanthopoulos, line 538: Maybe use structured binding and avoid copy while iterating

libtiledbvcf/src/write/writer_worker_v4.cc
```diff
@@ -157,16 +159,15 @@ bool WriterWorkerV4::resume() {
   //    fragment.
   //
   while (!record_heap_.empty()) {
-    RecordHeapV4::Node& top =
-        const_cast<RecordHeapV4::Node&>(record_heap_.top());
+    WriterRecordV4& top = const_cast<WriterRecordV4&>(record_heap_.top());
```
XanthosXanthopoulos, line 162: Although this is not introduced in this PR using `const_cast` should be avoided. This can be tackled in a future PR.

PR #880, opened 2026-04-01
Allow users to skip updating aggregate stats when deleting samples

alancleary:
Currently the delete samples code path is quite slow because all of the records for the samples being deleted need to be scanned so that the allele count and variant stats arrays can be updated. This PR adds a `--skip-aggregate-stats` flag to the `delete` CLI command and a `skip_aggregate_stats` parameter to `Dataset.delete_samples()` in the Python API to allow users to skip updating the allele count and variant stats arrays when deleting samples.

On a test dataset containing records from 50 VCFs, using this feature when deleting 2 samples changed the run-time from 22 minutes to less then 1 second.

### DIFF
--- a/libtiledbvcf/src/stats/allele_count.cc
+++ b/libtiledbvcf/src/stats/allele_count.cc
@@ -211,6 +211,41 @@ void AlleleCount::close() {
   enabled_ = false;
 }
 
+void AlleleCount::skip_delete_sample(
+    std::shared_ptr<Context> ctx,
+    const Group& group,
+    const std::string& sample) {
+  auto uri = AlleleCount::group_uri(group);
+  if (uri.empty()) {
+    return;
+  }
+
+  // Read existing skipped_delete_samples metadata
+  std::string existing;
+  {
+    Array read_array(*ctx, uri, TILEDB_READ);
+    const void* value = nullptr;
+    tiledb_datatype_t type = TILEDB_ANY;
+    uint32_t num = 0;
+    read_array.get_metadata("skipped_delete_samples", &type, &num, &value);
+    if (value != nullptr && num > 0) {
+      existing = std::string(static_cast<const char*>(value), num);
+    }
+  }
+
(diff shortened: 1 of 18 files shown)


leipzig:
can we make sure "double deletions" don't have unintended consequences, the second one should fail

spirpinias (approved):
great work, Alan. Much appreciate the haste!

leipzig (approved):

libtiledbvcf/src/stats/allele_count.cc
```diff
@@ -211,6 +211,41 @@ void AlleleCount::close() {
...
+
+  // Build updated CSV
+  std::string csv = existing.empty() ? sample : existing + "," + sample;
+
+  // Write updated metadata
+  {
+    Array write_array(*ctx, uri, TILEDB_WRITE);
+    write_array.put_metadata(
+        "skipped_delete_samples", TILEDB_STRING_ASCII, csv.size(), csv.c_str());
+  }
```
jp-dark: If multiple delete queries are kicked off at the same time, we could end up where the `skipped_delete_samples` from one delete query clobbers the value from another delete query.

libtiledbvcf/src/stats/variant_stats.cc
```diff
@@ -309,6 +309,41 @@ void VariantStats::close() {
...
+
+  // Build updated CSV
+  std::string csv = existing.empty() ? sample : existing + "," + sample;
+
+  // Write updated metadata
+  {
+    Array write_array(*ctx, uri, TILEDB_WRITE);
+    write_array.put_metadata(
+        "skipped_delete_samples", TILEDB_STRING_ASCII, csv.size(), csv.c_str());
+  }
```
jp-dark: Same comment as above about possibly over-writing metadata with concurrent writes.

XanthosXanthopoulos (approved):

PR #883, opened 2026-04-09
Support pybind11 3.0

kounelisagis:
Replace deprecated `py::module` with `py::module_` in Python bindings to support pybind11 3.0 while maintaining backward compatibility with 2.x.


alancleary (approved):
Thanks for jumping on this @kounelisagis. Looks good!

alancleary:
@kounelisagis I'm going to go ahead and merge so we can proceed with the next VCF release.

PR #896, opened 2026-08-05
Reverted TileDB VFS S3 retry strategy in the nightly workflow

alancleary:
The unreleased TileDB 2.31 (i.e. main) changes the VFS S3 retry strategy to match that of the AWS SDK's. This has caused the run-time of the TileDB-VCF nightly workflow to balloon on all runners, with the most egregious being the macOS runner. While the slowness in the workflow specifically came from the S3 read in `tests/test_config.py::test_large_export_correctness`, the problem is systemic and so the solution committed here is systemic as well.


jdblischak (commented):
Unfortunately this did not decrease the runtime of the pytest suite for the [macOS build against the main branch of TileDB](https://github.com/TileDB-Inc/TileDB-VCF/actions/runs/31020707282/job/92356361556?pr=896):

Image

jdblischak:
xref: https://github.com/TileDB-Inc/TileDB/pull/5747

alancleary:
This branch/PR has devolved into debugging. Closing and will follow-up with a fix in a clean branch

